PR #16, opened 2017-03-11
Fix gflags compilation issue

tiktuk:
Fixed compilation on Debian by changing the way gflags is imported.

(Sorry, I messed up the commits a bit, am not using Git usually.)


MisterTea:
Thanks!

PR #22, opened 2017-03-28
Portability

cgull:
This has various fixes for portability, correctness, and unused code that I discovered while trying to make ET work on FreeBSD.  I've removed most of the Linux- and OSX-isms, and it should be easy to get ET running on most POSIX OSes now.


launcher/et
```diff
@@ -32,7 +32,7 @@ do
             shift
             ;;
         -h|--help)
-            printf "et (options) [user@]hostname[:port]
+            printf "et (options) [user@]hostname
```
MisterTea: Why remove this?

src/CryptoHandler.cpp
```diff
@@ -29,7 +24,7 @@ CryptoHandler::~CryptoHandler() {}
 
 string CryptoHandler::encrypt(const string& buffer) {
   lock_guard<std::mutex> guard(cryptoMutex);
-  incrementNonce();
+  sodium_increment(nonce, crypto_secretbox_NONCEBYTES);
```
MisterTea: Cool!

launcher/et
cgull: * It conflicts with IPv6 literals, i.e. `mistertea@::1`
* There is code to parse out `:host` but the resulting variable isn't actually used anywhere.
* There's already a `-p` option to provide the info, the same way SSH does it.

cgull:
That's an old version of libsodium you've got in your Travis config-- it calls itself `1.0.2`, which doesn't exist on the libsodium project, but the project does show 1209 commits to master since 1.0.3.  You might consider upgrading it.

launcher/et
cgull: Another way to fix this would be to implement quoted IPv6 literals, i.e. `mistertea@[::1]`, but correctly parsing IPv6 literals is painful and best done with library functions rather than regexes.

cgull: Oops, I see the port info *is* properly propagated, and my commit does eliminate functionality.  Skip it if you don't like it :)

MisterTea: Your IPV6 argument makes sense.  I'll update the docs.

MisterTea:
Thanks so much for your work!

terminal/TerminalServer.cpp
```diff
@@ -59,18 +63,13 @@ void runTerminal(shared_ptr<ServerClientConnection> serverClientState,
...
     fd_set rfd;
-    fd_set wfd;
-    fd_set efd;
     timeval tv;
 
     FD_ZERO(&rfd);
-    FD_ZERO(&wfd);
-    FD_ZERO(&efd);
     FD_SET(masterfd, &rfd);
-    FD_SET(STDIN_FILENO, &rfd);
```
MisterTea: Why delete line 70?  Without this, we will wait for a timeout when there is data to send to the client immediately.

MisterTea:
cc: @begui Can you update the version of libsodium in travis?

terminal/TerminalServer.cpp
cgull: `etserver` never actually *reads* from fd 0, at least that I saw in strace or in your source.  (I wouldn't expect a network daemon running in background to have anything usefully connected on the standard file descriptors.)  It checks for console output from the user's session on `masterfd` here, with a timeout of 1ms, and then reads it in the block just below here.  It checks for network input (with a *different* `select()`) and reads it in `UnixSocketHandler::hasData()` with a timeout of 0.

With the original code, it's also checking for input on fd 0.  'select()` never waits and always returns a ready status from that file descriptor.  I didn't delve into why, but probably fd 0 is closed or otherwise returning an error, which will cause `select()` to report that fd as readable.  So the program consumes 100% CPU.

Removing this reference makes the program repeatedly sleep for 1ms in the first `select()` if nothing is available, polling the other `select()` each time.  This is still inefficient code, but far better than before.  This patch applies that simple band-aid.

Ideally you would use a single `select()` for the entire program and not have to time out so you can go poll some other `select()` and fds.  That, or convert the program to use an event handling framework.

MisterTea: Ah, you are right.  Thank you so much for your PR.  I also have a fix where the single select checks the fd in UnixSocketHandler but it wasn't giving me less CPU cycles on the server, and now I know why :-)

PR #26, opened 2017-04-11
Change SSH command to invoke '/bin/bash' on connection

cjthompson:
This allows et to connect to execute the password command on servers that do not have a BASH compatible shell as the default.

I don't feel like this is the *best* approach to this problem. However, it does work for me.  With a remote shell of `/usr/local/bin/fish`, the password command now succeeds.

It's possible that this will cause problems with the `--ssh=` parameter if the user adds a command rather than just flags.


MisterTea:
Thanks!

PR #44, opened 2017-06-26
AA: Fixing redundant forcing of current user and allowing ssh to pick username based on host

eskhool:
Tested and works with and without the username


MisterTea:
Thanks!

Issue #48, opened 2017-07-03
Improve documentation for using custom ssh commands.


mbtamuli:
My server listens on port `2222` instead of the standard `22`.
```
mbtamuli@nyc-512mb ~ » sudo ufw status numbered
Status: active

     To                         Action      From
     --                         ------      ----
[ 1] 22                         DENY IN     Anywhere                  
[ 2] 2222                       ALLOW IN    Anywhere                  
[ 3] 2022                       ALLOW IN    Anywhere                  
[ 4] 22 (v6)                    DENY IN     Anywhere (v6)             
[ 5] 2222 (v6)                  ALLOW IN    Anywhere (v6)             
[ 6] 2022 (v6)                  ALLOW IN    Anywhere (v6)             
```
I opened the `2022` port for et. Still I am not able to connect. (I already have open `2222`)
```
$ et mbtamuli@example.com:2222
ssh: connect to host techfreak.ga port 22: Connection timed out
SSH handshake failed.
```

I can confirm it's working perfectly fine for my other server which is listening on the standard SSH port `22`

MisterTea:
Et and ssh run on different ports. To change the ssh port, you need to pass
a flag to ssh itself using the -s command.

mbtamuli:
Can you please describe how to do that? It doesn't seem to be documented. I tried reading the launcher to figure it out, but from what I could see, the `et` bash launcher sends the port specified at the end of the whole command to the. The `:2222` at the end(As per line [#79](https://github.com/MisterTea/EternalTCP/blob/master/launcher/et#L79)). I may be wrong.

I've tried the following commands.
```
$ et -s="mbtamuli@example.com:2222" mbtamuli@example.com
ssh: Could not resolve hostname example.com:2222: Name or service not known
SSH handshake failed.
$ et -s="ssh -p 2222 mbtamuli@example.com" mbtamuli@example.com
ssh: Could not resolve hostname ssh: Name or service not known
SSH handshake failed.
$ et -s="ssh -p 2222" mbtamuli@example.com
ssh: Could not resolve hostname ssh: Name or service not known
SSH handshake failed.
$ et -s="-p 2222" mbtamuli@example.com
ssh: Could not resolve hostname /bin/bash: Name or service not known
SSH handshake failed.
$ et -s="2222" mbtamuli@example.com
ssh: connect to host 2222 port 22: Invalid argument
SSH handshake failed.
```

MisterTea:
[...]

PR #49, opened 2017-07-06
Add example for ssh command in launcher help

mbtamuli:
Partial fix for #48 

Signed-off-by: Mriyam Tamuli <mbtamuli@gmail.com>

### DIFF
--- a/launcher/et
+++ b/launcher/et
@@ -49,7 +49,7 @@ do
 Options:
 -h Basic usage
 -s Special ssh command if necessary, otherwise set to user@hostname:22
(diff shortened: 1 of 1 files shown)


launcher/et
```diff
@@ -49,6 +49,7 @@ do
 Options:
 -h Basic usage
 -s Special ssh command if necessary, otherwise set to user@hostname:22
+   example: et -s '-p 8888 hostname' user@hostname:8080 (et running on port 8080)
```
MisterTea: Ditto for this, please change to user@hostname

PR #50, opened 2017-07-06
Add usage example for non-standard ssh port

mbtamuli:
Partial fix for #48

Signed-off-by: Mriyam Tamuli <mbtamuli@gmail.com>

### DIFF
--- a/usermanual.md
+++ b/usermanual.md
@@ -15,6 +15,7 @@ title: Tutorial
 1. ET uses ssh for handshaking and encryption, so you must be able to ssh into the machine from the client.  Make sure that you can ```ssh user@hostname```.
 2. ET uses TCP, so you need an open port on your server.  By default, it uses 2022.
 3. Once you have an open port, the syntax is similar to ssh: ```et user@hostname[:port]```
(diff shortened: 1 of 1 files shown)


usermanual.md
```diff
@@ -15,6 +15,7 @@ title: Tutorial
 1. ET uses ssh for handshaking and encryption, so you must be able to ssh into the machine from the client.  Make sure that you can ```ssh user@hostname```.
 2. ET uses TCP, so you need an open port on your server.  By default, it uses 2022.
 3. Once you have an open port, the syntax is similar to ssh: ```et user@hostname[:port]```
+4. If you have SSH listening on a different port you can use, ```et -s="-p [SSHport] hostname" user@hostname[:ETport]```
```
MisterTea: Change to:
```
et -s="-p [SSHport] user@hostname" user@hostname[:ETport]
```

MisterTea:
Thanks!

usermanual.md
MisterTea: We moved to using ssh_config, so set the port in your ssh_config and use et as normal.

PR #61, opened 2017-09-05
Add port forward range in -t

ailzhang:
Here are some log entries after the change.
./et -t="8000-8003:21-24" server:8080
screen shot 2017-09-04 at 5 19 09 pm

./et -t="8000-8003:21-24, 8010-8012:9000-9002" server:8080
screen shot 2017-09-04 at 5 20 37 pm

### DIFF
--- a/terminal/TerminalClient.cpp
+++ b/terminal/TerminalClient.cpp
@@ -34,7 +34,7 @@ DEFINE_string(idpasskeyfile, "",
               "File containing client ID and key to encrypt/decrypt packets");
 DEFINE_string(command, "", "Command to run immediately after connecting");
 DEFINE_string(portforward, "",
-              "Array of source:destination ports or srcStart-srcEnd:dstStart-dstEnd (inclusive) port ranges (e.g. 10080:80,10443:443, 10090-10092:8000-8002)");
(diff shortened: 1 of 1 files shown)


terminal/TerminalClient.cpp
```diff
@@ -165,13 +165,36 @@ int main(int argc, char** argv) {
...
+	  } else {
+	    int portRangeLength = sourcePortEnd - sourcePortStart + 1;
+	    for (int i=0; i < portRangeLength; ++i) {
+	      portForwardRouter.addListener(
+		  shared_ptr<PortForwardClientListener>(new PortForwardClientListener(
+		      socketHandler, sourcePortStart+i, destinationPortStart+i)));
+	    }
+	  }
+	} else {
+          // TODO: Handle bad input
```
MisterTea: Can you also handle bad input as part of this PR?

All you should need to do is add catches for std::invalid_argument and std::out_of_range to the catch clause below.  Printing to cout and exit(1) should be sufficient.

Thanks!

ailzhang: Sure!

MisterTea: Actually if you want to only add 1 more catch, you can use std::logic_error which includes both of the ones I mentioned above.

PR #64, opened 2017-09-27
SSH config file parser

ailzhang:
Support parsing ssh config files for et to use. User-specific config file(~/.ssh/config) has higher priority than system-wide config file(/etc/ssh/ssh_config). 
Currently these options below are parsed. Hostname field is used to help translate et host field. 
For example, if you have a hostname alias `dev` for `someremotemachine.com`. you can now do `et dev:8080` instead of `et someremotemachine.com:8080`.
Support for other config options can be added based on request(such as jumphost).
```
Include (to include another config file)
Host
Hostname
port
username
connecttimeout
protocol
stricthostkeychecking
userknownhostsfile
proxycommand
GSSAPISeverIdentity
GSSAPIClientIdentity
GSSAPIDelegateCredentials
```

### DIFF
--- a/terminal/TerminalClient.cpp
+++ b/terminal/TerminalClient.cpp
@@ -7,6 +7,7 @@
 #include "ServerConnection.hpp"
 #include "SocketUtils.hpp"
 #include "UnixSocketHandler.hpp"
+#include "ParseConfigFile.hpp"
 
 #include <errno.h>
 #include <pwd.h>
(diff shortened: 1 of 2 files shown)


terminal/TerminalClient.cpp
```diff
@@ -135,6 +136,29 @@ int main(int argc, char** argv) {
...
+    NULL, //gss_server_identity
+    NULL, //gss_client_identity
+    0 //gss_delegate_creds
+  };
+  char* home_dir = ssh_get_user_home_dir();
+  ssh_options_set(&options, SSH_OPTIONS_HOST, FLAGS_host.c_str());
+  /* First parse user-specific ssh config, then system-wide config. */
+  parse_ssh_config_file(&options, string(home_dir) + "/.ssh/config");
+  parse_ssh_config_file(&options, "/etc/ssh/ssh_config");
+  cout << "connecting to " << options.host << endl;
```
MisterTea: Change to LOG(INFO).  If we want to print to cout, that's fine but let's also add a LOG(INFO) line so it gets logged to file as well.

ailzhang:
Fixed :) There are also some cout for error message before exiting in ParseConfigFile.hpp. I feel it's good to output sth before exiting, let me know if you would like them to be logged as well. Thanks!

MisterTea:
Thanks!

PR #67, opened 2017-10-17
Jumphost feature in ET

ailzhang:
Sorry this is a big patch :P. 
There are several things done in this PR.
---first commit: all -J related---
1. Start et from etclient directly instead of et script. 
2. -J jumphost feature
3. Fixed double close of server client connection
4. Added --user option in etclient to support connecting to different user accouts other than the default username. This was previously done in et script.
5. Updated readme.
---second commit: intergrate with sshconfig parser.--
6. Fixed LOG(INFO) printing unsupported keyword. 
7. format ParseConfigFile.hpp
Thanks for the help! Please let me know if you have any suggestions.


README.md
```diff
@@ -70,7 +70,13 @@ ET uses ssh for handshaking and encryption, so you must be able to ssh into the
 
 ET uses TCP, so you need an open port on your server. By default, it uses 2022.
 
-Once you have an open port, the syntax is similar to `ssh`: `et user@hostname[:port]`. You can pass additional arguments to `ssh` using the `-s` parameter. For instance, if you have `sshd` listening on port 5000: `et -s="-p 5000 user@host" user@host`.
```
MisterTea: Why is this comment gone?  Is the -s switch removed in this PR?

launcher/et
```diff
@@ -1,134 +0,0 @@
-#!/bin/bash
```
MisterTea: Please update the CMakeLists.txt file to rename the etclient binary to just "et" so people can continue to use the et command.

terminal/SocketUtils.hpp
```diff
@@ -1,10 +1,23 @@
 #include "Headers.hpp"
```
MisterTea: Let's rename this file to RawSocketUtils.hpp so it's more clear when to use this versus ET's sockethandler.  Also, let's create a class called RawSocketUtils and make these methods static methods of that class.

terminal/TerminalClient.cpp
```diff
@@ -130,7 +137,139 @@ void handleWindowChanged(winsize* win) {
   }
 }
 
-int main(int argc, char** argv) {
+void initSetupSSH() {
```
MisterTea: Let's move this into a file called SshSetupHandler.hpp/cpp and make initSetupSSH take the passkey, id, jumphost as parameters

terminal/TerminalClient.cpp
```diff
@@ -130,7 +137,139 @@ void handleWindowChanged(winsize* win) {
...
+  string CLIENT_TERM(getenv("TERM"));
+  if (FLAGS_passkey.empty()) {
+    FILE *passkey_p = popen(
+        "env LC_ALL=C tr -dc \"a-zA-Z0-9\" < /dev/urandom | head -c 32", "r");
+    if (!passkey_p) {
+      cout << "cannot generate passkey!" << endl;
```
MisterTea: LOG(FATAL)

terminal/TerminalClient.cpp
```diff
@@ -130,7 +137,139 @@ void handleWindowChanged(winsize* win) {
...
+    char passkey_buffer[1024];
+    fgets(passkey_buffer, sizeof(passkey_buffer), passkey_p);
+    pclose(passkey_p);
+    FLAGS_passkey = string(passkey_buffer);
+  }
+  if (FLAGS_id.empty()) {
+    FILE *id_p = popen(
+        "env LC_ALL=C tr -dc \"a-zA-Z0-9\" < /dev/urandom | head -c 16", "r");
+    if (!id_p) {
+      cout << "cannot generate id!" << endl;
```
MisterTea: LOG(FATAL)

terminal/TerminalClient.cpp
```diff
@@ -130,7 +137,139 @@ void handleWindowChanged(winsize* win) {
...
+
+
+  string cmdoptions = "--idpasskey=" + FLAGS_id + "/" + FLAGS_passkey;
+  string SSH_SCRIPT_PREFIX{
+      "SERVER_TMP_DIR=${TMPDIR:-${TMP:-${TEMP:-/tmp}}};"
+      "TMPFILE=$(mktemp $SERVER_TMP_DIR/et-server.XXXXXXXXXXXX);"
+      "PASSKEY="+FLAGS_passkey+";"
+      "ID="+FLAGS_id+";"
+      "export TERM="+CLIENT_TERM+";"
+      "/usr/bin/etserver "};
```
MisterTea: we can't assume that people have etserver installed to /usr/bin.  Does putting etserer by itself fail when connecting to os/x?  If so, we need to run the ssh commands in interactive mode so the bashrc/zshrc files are executed.

terminal/TerminalClient.cpp
```diff
@@ -130,7 +137,139 @@ void handleWindowChanged(winsize* win) {
...
+    }
+    cmdoptions +=
+        " --dsthost=" + FLAGS_host + " --dstport=" + to_string(FLAGS_port);
+  }
+  string SSH_SCRIPT_JUMP = SSH_SCRIPT_PREFIX + cmdoptions + ";true";
+
+  int link[2];
+  char foo[4096];
+  if (pipe(link) == -1) {
+    cout << "pipe" << endl;
```
MisterTea: LOG(FATAL).  There are more cout -> LOG(FATAL)s, so I won't list all of them

terminal/TerminalClient.cpp
```diff
@@ -130,7 +137,139 @@ void handleWindowChanged(winsize* win) {
...
+      string id = idpasskey_splited[0];
+      string passkey = idpasskey_splited[1];
+      if (id == FLAGS_id && passkey == FLAGS_passkey) {
+        LOG(INFO) << "etserver started" << endl;
+      } else {
+        LOG(INFO) << id << endl;
+        LOG(INFO) << FLAGS_id << endl;
+        LOG(INFO) << passkey << endl;
+        LOG(INFO) << FLAGS_passkey << endl;
+        cout << "ID " << FLAGS_id << endl;
```
MisterTea: Remove couts

terminal/TerminalClient.cpp
```diff
@@ -130,7 +137,139 @@ void handleWindowChanged(winsize* win) {
...
+        cout << "Received ID " << id << endl;
+        cout << "PASSKEY " << FLAGS_passkey << endl;
+        cout << "Received PASSKEY " << passkey << endl;
+        cout << "client/server idpasskey doesn't match" << endl;
+        exit(1);
+      }
+    } catch (const runtime_error &err) {
+      cout << "Error initializing connection" << err.what() << endl;
```
MisterTea: Is this a LOG(FATAL) or a LOG(ERROR)?  What should happen in this case?

terminal/TerminalClient.cpp
```diff
@@ -139,7 +278,7 @@ int main(int argc, char** argv) {
   FLAGS_logbuflevel = google::GLOG_INFO;
   srand(1);
 
-  Options options = {
+  Options o = {
```
MisterTea: Let's revert this back to putting options on the stack now that we fixed the other issue.

terminal/TerminalClient.cpp
```diff
@@ -130,7 +137,139 @@ void handleWindowChanged(winsize* win) {
...
+      exit(1);
+    }
+    char id_buffer[1024];
+    fgets(id_buffer, sizeof(id_buffer), id_p);
+    pclose(id_p);
+    FLAGS_id = string(id_buffer);
+  }
+
+
+  string cmdoptions = "--idpasskey=" + FLAGS_id + "/" + FLAGS_passkey;
```
MisterTea: idpasskey is not secure because command line options are visible to all users (ps aux).  The original et script put the passkey & id in a file and used the idpasskeyfile option which reads from that file and then destroys it.

terminal/TerminalClient.cpp
```diff
@@ -154,13 +293,18 @@ int main(int argc, char** argv) {
...
-  FLAGS_host = string(options.host);
+  ssh_options_set(options.get(), SSH_OPTIONS_HOST, FLAGS_host.c_str());
+  // First parse user-specific ssh config, then system-wide config.
+
+  parse_ssh_config_file(options.get(), string(home_dir) + USER_SSH_CONFIG_PATH);
+  parse_ssh_config_file(options.get(),SYSTEM_SSH_CONFIG_PATH);
+  LOG(INFO) << "Parsed ssh config file, connecting to " << options->host << endl;
+  FLAGS_host = string(options->host);
+
+  initSetupSSH();
```
MisterTea: SshHandler::initSetupSsh(...)

terminal/TerminalServer.cpp
```diff
@@ -358,7 +427,120 @@ void startUserTerminal() {
   uth.run();
 }
 
-int main(int argc, char** argv) {
+void startJumpHostClient() {
+  string idpasskey = FLAGS_idpasskey;
```
MisterTea: in startUserTerminal we handle either idpasskeyfile or commandline specified idpasskey.  Let's move that logic into main.  So, in main, if the idpasskeyfile option is set, we load the passkey from file and set FLAGS_idpasskey.  Then this function doesn't have to change and we don't have to check for FLAGS_idpasskeyfile before starting startUserTerminal

terminal/TerminalServer.cpp
```diff
@@ -55,13 +56,75 @@ DEFINE_string(idpasskeyfile, "",
...
+    }
+    tv.tv_sec = 0;
+    tv.tv_usec = 10000;
+    select(maxfd + 1, &rfd, NULL, NULL, &tv);
+
+    try {
+      if (FD_ISSET(terminalFd, &rfd)) {
+        memset(b, 0, BUF_SIZE);
+        try {
+          string message = readMessage(terminalFd);
```
MisterTea: This will change to RawSocketUtils::readMessage(...)

ailzhang:
Hi Jason, I just pushed a new commit to address the comments.
The main changes are:
1. Now ssh config also read ProxyJump (jumphost) from the file.
2. Security: use idpasskeyfile to send idpasskey
3. Fix LOG(INFO) issue in ssh config parser.
4. Update README
5. run format.sh
Please let me know if you have any suggestions.

terminal/TerminalClient.cpp
```diff
@@ -228,13 +235,13 @@ int main(int argc, char** argv) {
...
-          cout << "Logic error: " << lr.what() << endl;
+          LOG(FATAL) << "Logic error: " << lr.what() << endl;
           exit(1);
         }
       }
     }
   } catch (const std::runtime_error& ex) {
-    cout << "Error establishing port forward: " << ex.what() << endl;
+    LOG(FATAL) << "Error establishing port forward: " << ex.what() << endl;
     exit(1);
```
MisterTea: nit: remove this line, since it will exit after FATAL

terminal/TerminalServer.cpp
```diff
@@ -358,7 +427,130 @@ void startUserTerminal() {
...
+    std::ifstream t(FLAGS_idpasskeyfile.c_str());
+    std::stringstream buffer;
+    buffer << t.rdbuf();
+    idpasskey = buffer.str();
+    // Trim whitespace
+    idpasskey.erase(idpasskey.find_last_not_of(" \n\r\t") + 1);
+    // Delete the file with the passkey
+    remove(FLAGS_idpasskeyfile.c_str());
+  }
+  cout << "IDPASSKEY:" << idpasskey << endl;
```
MisterTea: remove this cout

ailzhang: Hey, this line was redirected to etclient for checking if ID & PASSKEY matches the one we have in the client side. It's similar to the cout logic in startUserTerminal. I feel we should keep it?

MisterTea: Oh!  You are totally right, it actually breaks without that cout :-P.  Thanks for catching that.

PR #69, opened 2017-10-29
Make compliation on FreeBSD work

telser:
The changes are really simple:

1. Don't need SELinux on FreeBSD (for obvious reasons), and compilation fails for some reason if find_package is called looking for it. So I added a guard around that.

2. FreeBSD doesn't need, and in fact can't use libresolv, with that functionality in libc. This meant two changes, one of which can **definitely** be done better. The first is to include sys/socket.h in the FreeBSD includes block. The second is to not link in libresolv. My experience which cmake is very limited, but this seemed to work for me.

Clearly the linking changes in `CMakeLists.txt` could be better, so whatever needs to be done for that just let me know.

Lastly, creating a port for this, so et can be available as a package in the standard set on FreeBSD, should be relatively straightforward and I'll gladly do that as well. Seemed better to not have a patch in there, so doing this part first.

EDIT:

Some further information. The changes were to get things to compile on FreeBSD-11-STABLE. And I didn't run the tests, (which link resolv) on FreeBSD, but I can make that change with further advice on the cmake changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,9 @@ find_package(Glog REQUIRED)
 find_package(GFlags REQUIRED)
 find_package(Protobuf REQUIRED)
 find_package(Sodium REQUIRED)
+IF(LINUX)
 find_package(SELinux)
+ENDIF()
 find_package(UTempter)
 
 IF(SELINUX_FOUND)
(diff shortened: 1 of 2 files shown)


telser:
A couple of further questions:

- Would a change to `README.md` detailing how to run tests, like travis.ci locally be acceptable? Perhaps it's just my general unfamiliarity with the combination of cmake and gmake used here, but it was non-obvious to me how to do that.

- If this is acceptable, would additional changes to get EternalTCP on pkgsrc, for use on NetBSD, (my personal second use case) illumos, and potentially others, be acceptable as well?

CMakeLists.txt
```diff
@@ -121,21 +123,38 @@ add_executable (
...
-  ${CMAKE_THREAD_LIBS_INIT}
-  ${PROTOBUF_LIBRARIES}
-  ${GLOG_LIBRARIES}
-  ${GFLAGS_LIBRARIES}
-  ${sodium_LIBRARY_RELEASE}
-  ${SELINUX_LIBRARIES}
-  ${UTEMPTER_LIBRARIES}
-  resolv
-  util
+IF(!FREEBSD)
```
MisterTea: Instead of putting the logic here, can you create a new cmake variable called CORE_LIBRARIES and set it to either resolv + util or just util depending on IF(FREEBSD)?

CMakeLists.txt
```diff
@@ -144,20 +163,36 @@ add_executable (
...
-  EternalTCP-static
-  ${CMAKE_THREAD_LIBS_INIT}
-  ${PROTOBUF_LIBRARIES}
-  ${GLOG_LIBRARIES}
-  ${GFLAGS_LIBRARIES}
-  ${sodium_LIBRARY_RELEASE}
-  ${UTEMPTER_LIBRARIES}
-  resolv
-  util
+IF(!FREEBSD)
```
MisterTea: If you follow the change above, you can also replace this IF by using ${CORE_LIBRARIES}

MisterTea:
W.R.T. your comments: Any changes to the README are gladly appreciated :-D

Also, creating a FreeBSD/NetBSD/etc. package would be great!  All I ask is that you submit a PR with a script that I can run to update the package so it doesn't go stale.

telser:
Changes requested should be done now. 

As for the packaging, I'm not sure how that script will work exactly, but will give it a shot. Will get on creating the ports.

CMakeLists.txt
```diff
@@ -19,6 +19,12 @@ if(UNIX)
   SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -std=c++11")
 endif()
 
+#Using FreeBSD?
+if (CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
+    set(FREEBSD TRUE)
+    set(BSD TRUE)
```
MisterTea: What's the point of setting BSD here?

telser:
Bah, didn't mean to keep that in there. Though it might make a comeback depending on what is needed on other platforms. Thanks for the quick turn around on this!

PR #74, opened 2017-12-10
Fix jump connection

ailzhang:
* Explicitly use Connection::closeSocket() to differentiate with the one in ClientConnection (which also contains reconnect)
* If jump client is idle, kill the jump->dst connection and restore when the src is back.
* Added some log to LOG(ERROR) due to the current log library issue.

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -10,7 +10,8 @@ Connection::~Connection() {
     LOG(ERROR) << "Call shutdown before destructing a Connection.";
   }
   if (socketFd != -1) {
-    closeSocket();
+    LOG(INFO) << "Connection destroyed";
+    Connection::closeSocket();
   }
 }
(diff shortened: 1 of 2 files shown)


src/Connection.cpp
```diff
@@ -30,7 +31,8 @@ ssize_t Connection::read(string* buf) {
   if (messagesRead == -1) {
     if (isSkippableError()) {
       // Close the socket and invalidate, then return 0 messages
-      LOG(INFO) << "Closing socket because " << errno << " " << strerror(errno);
+      LOG(ERROR) << "Closing socket because " << errno << " "
```
MisterTea: Let's change these back to LOG(INFO).  In the future we will ditch glog for spdlog or another one that does not have issues writing LOG(INFO) in forks

Issue #63, opened 2017-09-25
Too many logs at /tmp, can't disable them


teresaejunior:
Hello! I have noticed there are tens of */tmp/etserver** files in the server, and tens of */tmp/etclient** in the client. I tried running `et -l=/dev/null`, but it makes no difference.

Since `et` has been running mostly well, I would like to disable logs on both the server and client!

MisterTea:
Hey, FWIW you should only get one of these logs per et connection, which should happen very infrequently since the connection is supposed to last for weeks/months.

With that said, I agree that we should have a way of turning them off (a config param for etserver and a -s switch for etclient).

teresaejunior:
Yes, it's not that much of a problem, but it happens to me because I close the session whenever I finish what I was supposed to do.

teresaejunior:
Thank you!

PR #76, opened 2018-01-01
Fix log issues

ailzhang:
1. Before this PR all terminal specific logs are aggregated by user, stored in /tmp/etserver.{hostname}.{username}.log.INFO.{time}. We cannot tell the difference between log entries from different terminals of the same user. After the change, these logs are created per terminal session, named as /tmp/etserver_terminal_{ID}.INFO.{time}. This way it's much easier for debugging.
2. Add verbose level in /etc/et.cfg. 
Note that only et daemon started by systemd uses /etc/et.cfg, which means a new terminal/jumphost session will be created without reading this file and their VLOG level are not affected. We can definitely make them also read this section, although sounds a bit tricky. What do you think? @MisterTea

### DIFF
--- a/etc/et.cfg
+++ b/etc/et.cfg
@@ -6,3 +6,4 @@ port = 2022
 
 [Debug]
 verbose = 0
+silent = 0
--- a/etc/et.cfg
+++ b/etc/et.cfg
@@ -3,3 +3,6 @@
 
 [Networking]
 port = 2022
+
+[Debug]
+verbose = 0
--- a/proto/ETerminal.proto
+++ b/proto/ETerminal.proto
@@ -46,4 +46,5 @@ message InitialPayload {
 
 message ConfigParams {
   optional int32 vlevel = 1;
+  optional int32 minloglevel = 2;
 }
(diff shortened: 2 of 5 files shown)


MisterTea:
Ah I see the issue: we need the root process to tell the other processes to set the logging level.  Let me make an issue to track.

etc/et.cfg
```diff
@@ -3,3 +3,6 @@
 
 [Networking]
 port = 2022
+
+[Debug]
+verbose = 0
```
MisterTea: Since you have the context at the moment, can you also add a silent flag?  It can also be 0 by default, but with a value of 1 we do not log anything to file.  This can address #63

ailzhang: To me it makes more sense that "silent" means we only do log(ERROR) and ignore all INFO and WARNING. Or do we want to disable all the logs?

ailzhang: Never mind :) It definitely better to disable all logs in a "silent" mode.

MisterTea:
Cool!  This looks great!  Closes #63

PR #85, opened 2018-01-29
Compliation on NetBSD

telser:
This adds the necessary `#include`s and `#if`s for getting et up and going on NetBSD.

- NetBSD doesn't have all of the `ai_flags`
- Uses `freopen` and passes `stderr` as an arg, rather than assigning to it.
- Calls `waitpid` instead of `waitid`. NetBSD-7 doesn't have `waitid`. The manpage for NetBSD-CURRENT has `waitid` so presumably the OS specific bit can be removed once NetBSD-8 is released.
- This also removed some `#include`s that were in a conditional branch, but also unconditionally brought it.

In particular, I imagine the way `freopen` is handled is suboptimal. My underdeveloped C++ skillset is showing sadly. But this got everything to work. Any suggestions on improvement are appreciated.

### DIFF
--- a/src/Headers.hpp
+++ b/src/Headers.hpp
@@ -15,6 +15,10 @@
 #include <time.h>
 #include <unistd.h>
 
+#if __NetBSD__
+#include <util.h>
+#endif
+
 #include <algorithm>
 #include <array>
 #include <deque>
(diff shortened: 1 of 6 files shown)


MisterTea:
Hey @telser, how long have you been running ET on NetBSD?

telser:
@MisterTea about 2 weeks, but only as a client. If I can get things running on OpenBSD/sparc64 for the server then I'm very close to where mosh can go away for me. So that'll be my next project, if you're open to it.

And yes, my environment is super diverse. For et, my complete use case would be clients: NetBSD/amd64 and Linux/amd64 (specifically Void Linux) and servers: FreeBSD/amd64, OpenBSD/sparc64 and illumos/sparc64.

PR #86, opened 2018-01-29
Add -x to kill old sessions

ailzhang:
* Add -x to kill old sessions of the same user (excluding root)
* For ssh options like jumphost and username, use cmdline > sshconfig > local username(only for USER)


terminal/TerminalClient.cpp
```diff
@@ -43,6 +43,7 @@ DEFINE_string(t, "",
               "10080:80,10443:443, 10090-10092:8000-8002)");
 DEFINE_string(jumphost, "", "jumphost between localhost and destination");
 DEFINE_int32(jport, 2022, "port to connect on jumphost");
+DEFINE_bool(x, false, "flag to kill all old sessions belongs to the user");
```
MisterTea: "flag to kill all old sessions belonging to the user"

PR #88, opened 2018-01-31
Throw error in raw socket read/write

ailzhang:
* throw runtime error in raw socket read/write
* remove LOG(FATAL) for calls to raw socket read/write
* catch runtime error from raw socket read/write and LOG(FATAL) them except for runJumphost and runTerminal.
* close(routerFd) when jump client exit normally
* format code

### DIFF
--- a/terminal/TerminalServer.cpp
+++ b/terminal/TerminalServer.cpp
@@ -140,9 +140,8 @@ void runJumpHost(shared_ptr<ServerClientConnection> serverClientState) {
           string message = RawSocketUtils::readMessage(terminalFd);
           serverClientState->writeMessage(message);
         } catch (const std::runtime_error &ex) {
-          LOG(INFO) << "Terminal session ended";
+          LOG(INFO) << "Terminal session ended" << ex.what();
           run = false;
-          globalServer->removeClient(serverClientState->getId());
           break;
         }
       }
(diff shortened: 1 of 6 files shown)


terminal/TerminalServer.cpp
```diff
@@ -153,7 +153,13 @@ void runJumpHost(shared_ptr<ServerClientConnection> serverClientState) {
           if (!serverClientState->readMessage(&message)) {
             break;
           }
-          RawSocketUtils::writeMessage(terminalFd, message);
+          try {
+            RawSocketUtils::writeMessage(terminalFd, message);
+          } catch (const std::runtime_error &ex) {
+            LOG(INFO) << "Terminal session ended";
```
MisterTea: Let's log that it specifically ended because the unix socket died and also log ex.what()

terminal/TerminalServer.cpp
```diff
@@ -153,7 +153,13 @@ void runJumpHost(shared_ptr<ServerClientConnection> serverClientState) {
...
+            run = false;
+            globalServer->removeClient(serverClientState->getId());
```
MisterTea: This line shouldn't be necessary because setting run=false will cause line 316 to fire which does the same thing

ailzhang:
Testing on both dev and jumphost servers. Will report back after a few days :D

ailzhang:
Sorry for the delay. This PR is good to go. 
Started working on the logging refactor. Should be ready soon.

PR #92, opened 2018-02-20
Adds section to README.md for installing on FreeBSD

telser:
EternalTCP is now included in the official FreeBSD packages!


MisterTea:
Hey @telser , how does this work?  If I push a new release of ET, how can I update the FreeBSD package?

MisterTea:
Oh, I think I found it: https://svnweb.freebsd.org/ports/head/net/eternaltcp/

Can you rename to eternalterminal and also can you add a bash script to the PR that updates the package?  Thanks a lot

MisterTea:
Actually if "et" isn't taken, that would be an even better name

telser:
@MisterTea On FreeBSD the packages have a "maintainer" that is emailed when an outdated report is filed. The tree has its own set of committers and I am only a contributor there as well as here. Any issue with a port, including being outdated or a patch updating the version , is submitted via a bugzilla form at: https://bugs.freebsd.org/bugzilla/enter_bug.cgi?component=Individual%20Port%28s%29&product=Ports%20%26%20Packages

It does require login information to submit, but I could give a script automating that submission a shot.

Lastly, I had given the package the name "et" (in the "net" category like it is now at "net/eternaltcp"), but the upstream committers renamed to eternaltcp before merging the PR. :/

telser:
I'm more than happy to keep up with the releases here and build/test them on FreeBSD before submitting the patch to build the new release in the package tree.

MisterTea:
@telser cool, thanks a lot!  If possible, please change the name to "eternalterminal" (which will change this PR).  If it isn't possible to change, let me know and I can accept as is, but people will have trouble finding the package.

telser:
I've filed an issue/submitted a diff to get the name changed and the version bumped [here](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=226231).

MisterTea:
In case they wanted the name to match the GitHub repo name, I renamed the repo to EternalTerminal

yurivict:
@MisterTea In order to update the FreeBSD port, the port maintainer needs to submit a request on bugs.FreeBSD.org.

Yuri

MisterTea:
Great, thanks!

PR #95, opened 2018-02-28
Replace Glog with Easylogging

ailzhang:
* Removed GLOG, add easylogging 
* Max log file size set to 20MB.
* All log files go to /tmp
* User terminal / jumphost logs names include username for debugging
* Stack trace support enabled in easylogging
* For etserver, logs can be disabled by setting silent =1 in /etc/et.cfg
* For etclient, logs can be disabled by pass --silent from command line
* --v and --logtostdout are in both etclient and etserver, they used to replace the Macro in GLOG for debugging purpose.


telser:
@ailzhang something I've considered giving a shot is syslog support, particularly for `etserver`. Do you or @MisterTea have any thoughts on this? Won't spend my time on it, if you guys are very against it or something.

FWIW, it appears that easylogging should make this pretty easy, but no experience with the library.

terminal/TerminalClient.cpp
```diff
@@ -165,6 +168,39 @@ vector<pair<int, int>> parseRangesToPairs(const string& input) {
 }
 
 int main(int argc, char** argv) {
+  // easylogging parse verbose arguments, see [Application Arguments]
+  // in https://github.com/muflihun/easyloggingpp/blob/master/README.md
+  START_EASYLOGGINGPP(argc, argv);
```
MisterTea: Let's create a LogHandler.h/.cpp and move this macro, the INITIALIZE_EASYLOGGINGPP macro, and the config below into those files.

terminal/TerminalServer.cpp
```diff
@@ -534,14 +522,28 @@ void startJumpHostClient() {
...
-  FLAGS_logbufsecs = 0;
-  FLAGS_logbuflevel = google::GLOG_INFO;
-  srand(1);
+  // easylogging parse verbose arguments, see [Application Arguments]
+  // in https://github.com/muflihun/easyloggingpp/blob/master/README.md
+  START_EASYLOGGINGPP(argc, argv);
+  // GFLAGS parse command line arguments
+  ParseCommandLineFlags(&argc, &argv, false);
+
+  // Easylogging configurations
```
MisterTea: Move this into LogHandler and pass the filename, verbosity, etc. as arguments to some function in LogHandler

ailzhang:
@telser we are good with it as long as it's can be turn on/off in /etc/et.cfg and the syslog is an addition to the current logging files. :)

PR #104, opened 2018-04-12
setup easylogging log rotation-20M

ailzhang:
- Enable log rotation in easylogging.
- Set the max log file to 20MB, log file will be deleted when exceeding size limit.


ailzhang:
* Added an option in et.cfg to set the max logfile size by user(default is 20M). This will make debugging flexible. Etclient doesn't read et.cfg so its limit remains 20M.

Issue #109, opened 2018-05-04
Can't find etserver during initial connection


zefei:
I'm a first time user of et and is currently trying to use v4.2.1. During connection (after ssh password), et returns with error:
    bash: etserver: command not found
    Error:  The Eternal Terminal daemon is not running.  Please (re)start the et daemon on the server.

Doing `which etserver` on my server, I got `/usr/local/bin/etserver`.

ailzhang:
@zefei , could you double check if there's a etserver daemon running? `ps aux | grep etserver`

zefei:
I tried on two servers, one has etserver daemon running `/usr/local/opt/et/bin/etserver --cfgfile /usr/local/etc/et.cfg --daemon` and one has not. Connecting either server results in the same error `bash: etserver: command not found`.

MisterTea:
@zefei Latest master now runs the etserver in interactive mode.

zefei:
@MisterTea @ailzhang thank you for the quick fix!

phord:
Gah!  ssh shouldn't run in interactive mode unless you are running an interactive shell.  You are not.

Interactive mode has many other implications, including which shell is started, what profiles are loaded, etc.  A better option would be to include a config option specifying the server location, a client-side switch to let me specify the full path on the remote, or some intelligent search/retry option on the remote.

PR #110, opened 2018-05-05
start etserver in interactive mode

ailzhang:
This PR should fix issue #109.  I only tested the scenario without -J (temporary limited resources) and it works. Will update when I have more resources to test -J.


MisterTea:
Thanks!!

MisterTea:
Whoops.  @ailzhang I had to revert this because the "-lic" only works for bash.  I think we can pass the "-i" flag to the ssh command itself and that should be portable across shell programs.

MisterTea:
@ailzhang disregard that.  Actually "zsh -lic" works fine, even though the -l isn't in the documentation (at least I couldn't find it).  I reverted the revert.

PR #113, opened 2018-05-16
First version of Htm

MisterTea:


ailzhang:
Hi, thanks for the great patch! I've tried it on my mac, as I typed 'htm' a new terminal showed up and everything works fine inside. One nit I noticed is that, I'm not sure how to safely exit that window while preserve window state, so I tried 'ESC' (as in tmux CC, not working) and then ctrl-D. This caused terminal to freeze and every time I tried to restart 'htm', it shows the following error. 
screen shot 2018-05-16 at 7 02 47 pm

MisterTea:
Oh, good find!  The way to exit is to close the master tab, but I think if you try to type in that tab, it will mix with the commands going to htm and cause issues.  Let me make that UX better.

PR #119, opened 2018-05-31
Link with execinfo on FreeBSD because easylogging needs it

telser:
Just a tiny fix to get things compiling again on FreeBSD.

As soon as a new version is released I'll update the FreeBSD package.

Also, I've setup some smoke-test CI for each of FreeBSD, NetBSD, and OpenBSD. These weren't running the tests, but seeing the CircleCI setup, I'll be switching to that soon. If there is interest I can expose the status of these, or the logs. It's a pretty simple buildbot setup so far, but whatever would be helpful I can make accessible.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,11 @@ if (CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
     set(FREEBSD TRUE)
 endif (CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
 
+#Using NetBSD?
+if (CMAKE_SYSTEM_NAME MATCHES "NetBSD")
+  set(NETBSD TRUE)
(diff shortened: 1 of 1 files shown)


telser:
@MisterTea @ailzhang I've updated this PR to build on NetBSD too.

MisterTea:
@telser looks great, thanks!  Can you add a FreeBSD section to our circleci config as part of this PR?  It should be a copy-paste of our Ubuntu job with a different docker image. 

Thanks

telser:
@MisterTea Docker depends on the Linux kernel, so I can't do that :( Docker for macOS gets around this by just running a Linux vm in the background. There is a port of runc to FreeBSD, but to my knowledge that hasn't hit 1.0 on Linux yet either and CircleCI doesn't seem to support it anyway.

MisterTea:
Thanks!

PR #126, opened 2018-08-06
Update README.md with debian instruction for installation

albuic:
Update the documentation in README.md using the documentation in gh-pages to make it clear that a debian repo exist for github users.
Update the "install from source" for debian.


MisterTea:
Thanks!!!

PR #127, opened 2018-08-07
Fix reconnect issue with jumphost

ailzhang:
Have been testing this branch for a while and it's consistently more stable.
For the records, this PR contains a few features from @MisterTea and me, including 
* fix initial connection with -J
* make handling connection multi-threaded
* rename the log files to have seconds in the name.
* log thread name in easy logging to help debugging
* bug fixes for reconnect within jumphost context.


ailzhang:
I will add another commit to format the verbose logging later today.

PR #134, opened 2018-09-12
Add support for command prefix before etterminal/etserver commands on…

kecsap:
… the server side

This method is usable if the target system needs to run etterminal/etserver with
different user priviledges compared to the ssh user. It is useful if systemd kills
et processes similar to mosh:
https://github.com/mobile-shell/mosh/issues/399
https://github.com/mobile-shell/mosh/issues/730

Example: et -prefix "sudo systemd-run --scope sudo -u MY_USER"

It will execute e.g. "sudo systemd-run --scope sudo -u MY_USER etterminal" on the server side.

### DIFF
--- a/src/terminal/SshSetupHandler.cpp
+++ b/src/terminal/SshSetupHandler.cpp
@@ -24,7 +24,7 @@ string genRandom(int len) {
 }
 
 string genCommand(string passkey, string id, string CLIENT_TERM, string user,
-                  bool kill, string options) {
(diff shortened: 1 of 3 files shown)


MisterTea:
Cool, nice PR @kecsap !  Maybe we should use --prefix instead of -C?  We already have a -c and people might mix them up.

kecsap:
Ok, I updated!

PR #142, opened 2018-10-02
Fix unit tests

MisterTea:
Sadly this didn't expose any bug in ET to fix, but the unit tests are passing now.

### DIFF
--- a/src/base/Connection.cpp
+++ b/src/base/Connection.cpp
@@ -76,7 +76,7 @@ bool Connection::write(const string& buf) {
   if (bwws == BackedWriterWriteState::WROTE_WITH_FAILURE) {
     VLOG(4) << "Wrote with failure";
     // Error writing.
-    if (!errno) {
+    if (socketFd == -1) {
       // The socket was already closed
       VLOG(1) << "Socket closed";
     } else if (isSkippableError(errno)) {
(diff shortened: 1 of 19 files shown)


src/base/Connection.cpp
```diff
@@ -16,80 +16,46 @@ Connection::~Connection() {
...
-  int num_trails = CLIENT_timeout * 10;
-  for (int trials = 0; trials < num_trails; trials++) {
-    ssize_t messagesRead = reader->read(buf);
-    if (messagesRead == -1) {
-      if (trials < (num_trails - 1) && errno == EAGAIN) {
-        // If we get EAGAIN, assume the kernel needs to finish
-        // flushing some buffer and retry after a delay.
-        usleep(100000);
-        LOG(INFO) << "Got EAGAIN, waiting 100ms...";
-      } else if (trials == (num_trails - 1) && errno == EAGAIN) {
```
ailzhang: This was added to address the EAGAIN case which sometimes show up in the reconnect. I'm not sure if our refactor resolves this issue. 

I will run this branch in my environment to test after this PR is merged. :)

MisterTea: I think I figured out what was happening: BackedReader did not have the same locks that BackedWriter had, so it's possible to have a race condition.  The global "errno" variable just has the most recent error so if we ever run two unix commands at the same time, we risk clobbering it.  That's why we have to be so strict with the mutexes :-)

PR #173, opened 2019-02-20
build from source debian

eberna:
as suggested i modified the readme to include the instruction for building for Debian/Ubuntu (systemd service and config file).


MisterTea:
Thanks!

PR #207, opened 2019-07-25
Switch to using the protobuf-lite library

MichaelBuckley:
Hi,

This pull requests switches from using the full protobuf library to the lite version, which reduces our binary size by about 10 MB.

The `DebugString()` method was the only full protobuf feature being used in the ET code. It was being used in a few logs. I was able to provide a suitable replacement for the log line in which we know the message type, but the other two log lines are in templated methods. The best I could do there was to replace the debug string with the message type.

If this will make diagnosing issues too difficult, I could look into perhaps `#ifdef` ing the code into lite and regular variants.

Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -165,7 +165,7 @@ target_link_libraries (
   TerminalCommon
   et-lib
   ${CMAKE_THREAD_LIBS_INIT}
-  ${PROTOBUF_LIBRARIES}
+  ${PROTOBUF_LITE_LIBRARIES}
   ${sodium_LIBRARY_RELEASE}
   ${SELINUX_LIBRARIES}
   ${UTEMPTER_LIBRARIES}
@@ -183,7 +183,7 @@ target_link_libraries (
   TerminalCommon
   et-lib
   ${CMAKE_THREAD_LIBS_INIT}
-  ${PROTOBUF_LIBRARIES}
+  ${PROTOBUF_LITE_LIBRARIES}
   ${sodium_LIBRARY_RELEASE}
   ${SELINUX_LIBRARIES}
   ${UTEMPTER_LIBRARIES}
@@ -201,7 +201,7 @@ target_link_libraries (
     TerminalCommon
     et-lib
     ${CMAKE_THREAD_LIBS_INIT}
-    ${PROTOBUF_LIBRARIES}
+    ${PROTOBUF_LITE_LIBRARIES}
     ${sodium_LIBRARY_RELEASE}
     ${UTEMPTER_LIBRARIES}
     ${CORE_LIBRARIES}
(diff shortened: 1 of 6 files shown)


MisterTea:
Looks good, thanks!

prasant94:
Hi,

When I try to run ```cmake ../``` as per the install instructions, I get the follow error. Does it have something to do with this change?

```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
Protobuf_LITE_LIBRARY
    linked by target "etterminal" in directory /home/chidella/lib/EternalTerminal
    linked by target "et" in directory /home/chidella/lib/EternalTerminal
    linked by target "etserver" in directory /home/chidella/lib/EternalTerminal
    linked by target "htmd" in directory /home/chidella/lib/EternalTerminal
    linked by target "et-test" in directory /home/chidella/lib/EternalTerminal
    linked by target "htm" in directory /home/chidella/lib/EternalTerminal

-- Configuring incomplete, errors occurred!
See also "/home/chidella/lib/EternalTerminal/CMakeFiles/CMakeOutput.log".
See also "/home/chidella/lib/EternalTerminal/CMakeFiles/CMakeError.log".
```

I'm referring to the instructions to install and build from source.

MichaelBuckley:
Hi,

This issue is definitely related to this change. (Though it appears the instructions need updating. You may face other issues after fixing this issue. Because of this, I've been using the `build_static.sh` script to build ET. But of course, this will build static binaries that won't link against your OS's shared libraries.

In addition to the dependencies listed in the instructions, you'll need to get the protobuf-lite library installed. I believe in Ubuntu's apt repository, this library is named `libprotobuf-lite17`. You'll also need to make sure you have an up-to-date version of CMake.

prasant94:
In case anyone else is running into the same issue, here's how I got it to work.

```
sudo yum install protobuf-lite
cmake \
    -DProtobuf_LITE_LIBRARY_DEBUG=/usr/lib64/libprotobuf-lite.so.6 \
    -DProtobuf_LITE_LIBRARY_RELEASE=/usr/lib64/libprotobuf-lite.so.6 \
    ../

cd ../
make
sudo make install
```

Still haven't run the whole thing. But atleast I was able to get past the cmake command with this.

Hopefully the install instructions will be fixed soon.

MisterTea:
Can you update the install instructions and submit a PR?

MichaelBuckley:
FWIW, I got an Ubuntu VM up and running today, and I was able to compile using the instructions in the readme. It looks like the `libprotobuf-dev` package installs protobuf lite, and CMake was able to find it.

@prasant94 Which distribution are you having issues with? I'll be happy to update the instructions if I can figure it out.

PR #209, opened 2019-07-29
Write a pidfile for the 2nd child process for systemd.

jshort:
DaemonCreator forks twice which confuses systemd, so this changes has
the main child process write a pidfile which can be referenced in the
et.service systemd configuration.

### DIFF
--- a/src/base/DaemonCreator.cpp
+++ b/src/base/DaemonCreator.cpp
@@ -1,7 +1,7 @@
 #include "DaemonCreator.hpp"
 
 namespace et {
-int DaemonCreator::create(bool parentExit) {
+int DaemonCreator::create(bool parentExit, string childPidFile) {
   pid_t pid;
 
   /* Fork off the parent process */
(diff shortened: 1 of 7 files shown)


src/base/DaemonCreator.cpp
```diff
@@ -35,6 +35,20 @@ int DaemonCreator::create(bool parentExit) {
...
   if (pid > 0) exit(EXIT_SUCCESS);
 
+  /* Child process, write pid file */
+  if (childPidFile != "") {
+    int pidFilehandle = open(childPidFile.c_str(), O_RDWR|O_CREAT, 0600);
+    if (pidFilehandle == -1) {
+      LOG(FATAL) << "Error opening pidfile for writing: "<< childPidFile;
+    }
+
+    // Max pid length for x86_64 is 2^22 ~ 4000000
```
MisterTea: Let's make this char array huge (like 4096) or use stringstream so we don't have to worry about overflow.

jshort: Updated, see new commit.

MisterTea:
Thanks!!

PR #217, opened 2019-08-14
iOS: Use remove() instead of system() to remove log files

MichaelBuckley:
Hi again,

This is a change that allows libet to be compiled on iOS. The only change is that instead of calling `system()` to remove a file using `rm`, it uses the C++ standard `remove()` function.

I am unaware of any platform that EternalTerminal supports that has `rm` but not `remove()`, but I am more than happy to make changes if this causes a compatibility problem.


MisterTea:
Thanks!

Issue #211, opened 2019-07-31
Super weird behavior when connecting via etserver vs. ssh


dandv:
Any idea how come that the same commands run over an `et` connection behave differently vs. when they're run over a regular ssh shell connection?

1. [`crontab -e`](https://unix.stackexchange.com/questions/533012/crontab-waitpid-failed-waiting-for-pid-xxx-from-usr-bin-sensible-editor-n) produces an infinite loop of "crontab: waitpid() failed waiting for PID 3910 from "/usr/bin/sensible-editor": No child processes" messages
2. `apt update` (logged in as root) produces "Waited for apt-key but it wasn't there" for every repo:

  ```
# apt update
Hit:1 http://mirrors.digitalocean.com/ubuntu bionic InRelease
Hit:2 http://mirrors.digitalocean.com/ubuntu bionic-updates InRelease                                                                                                  
Hit:3 https://repos.influxdata.com/ubuntu bionic InRelease                                                                                                             
Hit:4 https://deb.nodesource.com/node_12.x bionic InRelease                                                                                                            
Hit:5 http://mirrors.digitalocean.com/ubuntu bionic-backports InRelease                                                                                                
Err:1 http://mirrors.digitalocean.com/ubuntu bionic InRelease                                                                                                          
  Waited for apt-key but it wasn't there
Hit:6 http://ppa.launchpad.net/git-core/ppa/ubuntu bionic InRelease                                                                                                    
Hit:7 http://security.ubuntu.com/ubuntu bionic-security InRelease                              
Err:2 http://mirrors.digitalocean.com/ubuntu bionic-updates InRelease    
  Waited for apt-key but it wasn't there
```

Might there be some missing environment variable that `et` connections don't load/set?

MisterTea:
Hey @dandv !  Can you run the "env" command on ssh and et sessions to see if there are any differences?  Once we see differences, then we can grep for them on your computer to figure out what setup ET is missing.

dandv:
Duh, should've thought of that :)

Here's the diff (all other vars are the same):

![image](https://user-images.githubusercontent.com/33569/62268005-f5933180-b3f3-11e9-851a-a128f26fbdd7.png)

MisterTea:
Hmm, those envars shouldn't cause the errors you are seeing?

You probably want to log in as a user and then "sudo su -" to get root.  Logging in as root can likely have issues.

MisterTea:
[...]

PR #218, opened 2019-08-15
Replace wget/unzip with git --recurse-submodules

dandv:
Per https://github.com/MisterTea/EternalTerminal/issues/211#issuecomment-519198645


MisterTea:
Thanks!

PR #223, opened 2019-08-22
Fix fatal log errors caused when iOS closes background connections

MichaelBuckley:
Hi,

These are a couple small changes which prevent calling `FATAL_FAIL` when iOS terminates the connection of an app that's suspended in the background.

The `isSkippableError()` function is called when reading or writing to/from a connection. If the error is skippable, the code assumes the connection has failed and calls `closeSocketAndMaybeReconnect()`. It seems safe to me to add `ENOTCONN` to the list of skippable errors in this context.

The other change prevents calling `close()` on an invalid file descriptor.

### DIFF
--- a/src/base/UnixSocketHandler.cpp
+++ b/src/base/UnixSocketHandler.cpp
@@ -159,7 +159,9 @@ void UnixSocketHandler::close(int fd) {
   auto m = it->second;
   lock_guard<std::recursive_mutex> guard(*m);
   VLOG(1) << "Closing connection: " << fd;
-  FATAL_FAIL(::close(fd));
+  if (fcntl(fd, F_GETFD) != -1) {
+    FATAL_FAIL(::close(fd));
+  }
   activeSocketMutexes.erase(it);
 }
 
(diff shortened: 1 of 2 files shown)


src/base/UnixSocketHandler.cpp
```diff
@@ -159,7 +159,9 @@ void UnixSocketHandler::close(int fd) {
   auto m = it->second;
   lock_guard<std::recursive_mutex> guard(*m);
   VLOG(1) << "Closing connection: " << fd;
-  FATAL_FAIL(::close(fd));
+  if (errno != EBADF && errno != EPIPE) {
```
MisterTea: errno is based on the most recently called system function, which is undefined in this case because close hasn't been called yet.  Can you provide another way to check if the fd is valid?

MichaelBuckley: Yes, you're right. In my case, close() was being called after a failed read(), but there are definitely code paths that don't call system functions before the close. I have updated the pull request to use fcntl() to check for a valid FD. I only check the return value of fcntl(), because in testing on iOS devices, I was able to get it to fail with at least 5 different errno values, depending on when and how the connection was terminated.

PR #226, opened 2019-09-01
Forward unix sockets & implement ssh agent forwarding

MisterTea:


src/terminal/forwarding/PortForwardHandler.cpp
```diff
@@ -29,13 +32,45 @@ void PortForwardHandler::update(vector<PortForwardDestinationRequest>* requests,
...
+      throw runtime_error("Do not set a source when forwarding named pipes");
+    }
+    SocketEndpoint source;
+    if (pfsr.has_source()) {
+      source = pfsr.source();
+    } else {
+      // Make a random file to forward the pipe
+      string sourcePattern = string("/tmp/et_forward_sock_XXXXXX");
+      string sourceDirectory = string(mkdtemp(&sourcePattern[0]));
+      ::chmod(sourceDirectory.c_str(), 0777);
```
jtakkala: @MisterTea  shouldn't this be 0700?

MisterTea: The root user creates the socket and then hands it off to the actual user (check for chmod).

But now that you mention it, it should be possible to have the user process create the socket, I'll look into it.

Issue #184, opened 2019-04-19
git clone crashes


vegesm:
Git clone fails when running from et:
```
$ git clone https://github.com/CMU-Perceptual-Computing-Lab/panoptic-toolbox
Cloning into 'panoptic-toolbox'...
remote: Enumerating objects: 22, done.
remote: Counting objects: 100% (22/22), done.
remote: Compressing objects: 100% (16/16), done.
remote: Total 241 (delta 6), reused 22 (delta 6), pack-reused 219
Receiving objects: 100% (241/241), 6.99 MiB | 1.55 MiB/s, done.
Resolving deltas: 100% (83/83), done.
error: waitpid for index-pack failed: No child processes
fatal: index-pack failed
error: waitpid for fetch-pack failed: No child processes
```
The same witpid error is raised when trying to commit. Using normal ssh, everything work fine. My et version is 5.1.9

MisterTea:
Hmm, I git clone all the time and haven't seen this.  What OS are you running on the server?

vegesm:
It's ubuntu 16.04, kernel: 4.4.0-141-generic

MisterTea:
Is this a home machine or are you using a public cloud?  It sounds like some SELinux restrictions

vegesm:
It is a workplace computer, SElinux is turned off but I run etserver in user space.

I did a bit of digging around and it seems the main culprit is that ET somehow disables SIGCHLD. When running trap in et:
```
$ trap
trap -- '' SIGCHLD
trap -- '' SIGTSTP
trap -- '' SIGTTIN
trap -- '' SIGTTOU
```
In ssh:
```
$ trap
trap -- '' SIGTSTP
trap -- '' SIGTTIN
trap -- '' SIGTTOU
```

According to strace git uses clone and wait4 to fork a subprocess and wait for it to finish. If the SIGCHLD handler is set to ignore then wait4 reports ECHILD ([link](https://linux.die.net/man/2/waitpid)). If I run the following code, everything works fine:
```C
#include <unistd.h> 
#include <signal.h>

int main() 
{ 
    signal(SIGCHLD, SIG_DFL);  // Reset SIGCHLD handler to default
    execl("/usr/bin/git", "git", "clone", "path_to_repo", (char*)NULL);
    return 0;
}
```
Interestingly, in ET `trap` does not work on SIGCHLD, whatever I try, the SIGCHLD action is always ignore. Using ssh, I can change the signal action without any problem.

[...]

PR #231, opened 2019-09-22
reset SIGCHLD to SIG_DFL

nitzan-shaked:
... just before exec'ing $SHELL, so child processes of the shell receive
SIGCHLD, even if the shell is not smart enough to do this itself (bash,
e.g.).


MisterTea:
This fixes https://github.com/MisterTea/EternalTerminal/issues/184

PR #233, opened 2019-09-24
Explain where to change the listening port. Fix #232.

### LINKED ISSUE #232: Document how to change the default port
I don't see in the README how to change the default port that `et` is listening on, from 2022 to something else.

dandv:


MisterTea:
Thanks!

PR #242, opened 2019-10-31
Describe backing connection w encryption

dtgriscom:
Adds description of backing connection, and that it is encrypted.


MisterTea:
Thanks!

Issue #197, opened 2019-07-18
SSH config "Include" path needs to be absolute path


grakkpl:
ET does parse the "Include" parameter from ssh config file but it assumes it is an absolute path, and SSH supports both - absolute and relative paths. 

Working example:
```
cat .ssh/config
Include /Users/barteq/.ssh/config.eda
```

Invalid example:
```
cat .ssh/config
Include config.eda
```

Both work through SSH, ET reports this:
`Error starting ET process through ssh, please make sure your ssh works first`

MisterTea:
@grakkpl Is this still an issue now that #244 has landed?

PR #244, opened 2019-11-04
Expand tilde before parsing include file

oyvinmar:
Adds support for tilde in included config files (which is supported by ssh).  Example:
```
Include ~/.ssh/conf/settings.ssh
Include ~/.ssh/conf/hosts.ssh
```
Related to https://github.com/MisterTea/EternalTerminal/issues/197

Severely lacking any cpp skills so let me know if I should change something.


MisterTea:
Looks good, thanks!!

PR #246, opened 2019-11-08
Add a build option to disable easylogging's default crash handler

MichaelBuckley:
Hi again. It's been a while.

We recently found that easylogging's default crash log installs a signal handler that interferes with our own signal handler when linking `libet`. Thankfully, it's fairly easy to disable, so I figured I would offer an upstream pull request for a build option to disable it.

Thanks for considering it!


MisterTea:
Thanks!  I assume no news is good news: glad to hear ET is working for you :-D

PR #258, opened 2019-11-27
Update README.md; add Raspbian build instructions

LaurentFough:
Add Raspbian build instructions
- install dependencies: `python3-protobuf protobuf-compiler`
- correct the`/etc/systemd/system/et.service` file to point to `/usr/local/bin/etserver`

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ sudo systemctl start et.service
 ```
 
 ### Raspbian (Stretch):
-## (_tested on_: Raspberry Pi 3B+):
+#### (_tested on_: Raspberry Pi 3B+):
 
 + Install dependencies
 
--- a/README.md
+++ b/README.md
@@ -230,6 +230,36 @@ sudo systemctl enable et.service
 sudo systemctl start et.service
 ```
 
+### Raspbian (Stretch):
+## (_tested on_: Raspberry Pi 3B+):
+
++ Install dependencies
+
+	sudo apt install python3-protobuf protobuf-compiler
+
++ Install from source
+```
+mkdir ~/tmp
+cd ~/tmp
+git clone --recurse-submodules https://github.com/MisterTea/EternalTerminal.git
+cd EternalTerminal
+mkdir build
+cd build
+cmake ../
+make
+sudo make install
+```
+
+Find the actual location of et:
+
+	which etserver
+
+Correct the service file:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -230,6 +230,36 @@ sudo systemctl enable et.service
 sudo systemctl start et.service
 ```
 
+### Raspbian (Stretch):
+#### (_tested on_: Raspberry Pi 3B+):
+
++ Install dependencies
+
+	sudo apt install python3-protobuf protobuf-compiler
```
MisterTea: I don't think we need python3, installing protobuf-dev should be good enough

MisterTea:
Hey @LaurentFough can you confirm that replacing python3-protobuf with protobuf-dev works?

README.md
qiuqiyuan: I was able to get enough dependency by running:
```
sudo apt-get install libprotobuf-dev
```

qiuqiyuan: On top of the protobuf deps, I also need to install:
```
sudo apt get libsodium-dev
sudo apt get libevent-pthreads-2.1-6
```

qiuqiyuan: My device is Raspberry Pi 4B with Raspbian.

README.md
```diff
@@ -230,6 +230,36 @@ sudo systemctl enable et.service
...
+## (_tested on_: Raspberry Pi 3B+):
+
++ Install dependencies
+
+	sudo apt install python3-protobuf protobuf-compiler
+
++ Install from source
+```
+mkdir ~/tmp
+cd ~/tmp
```
qiuqiyuan: nit: Maybe we do not need to create `tmp` folder here?
Reason being `git clone` will create a project folder by default, as result we can omit this step.

README.md
```diff
@@ -230,6 +230,36 @@ sudo systemctl enable et.service
...
+git clone --recurse-submodules https://github.com/MisterTea/EternalTerminal.git
```
qiuqiyuan: Thanks for creating this PR, I was struggling to find the right way to install ET.
From your PR I was able to learn that offering `--recurse-submodules` is a key argument because otherwise Raspbian does not have the compatible package ET depends on.

Pulling the correct source is the key to installing this successfully locally. Thank you!

MisterTea:
Raspbian should be supported using the Debian instructions so I'm going to close this.

PR #260, opened 2019-12-02
Mention that libprotobuf-dev is also required (by the client)

dandv:
Without libprotobuf-dev installed, on Ubuntu I got `error while loading shared libraries: libprotobuf-lite.so.10: cannot open shared object file: No such file or directory`. After installing it, `et` started fine.


MisterTea:
Thanks!

Issue #262, opened 2019-12-03
Add support for IdentityAgent and ProxyCommand for ssh agent forwarding


isaacs:
I use [Krypton](https://krypt.co/) to secure my ssh keys via 2-factor auth.  This puts the following entry in my ~/.ssh/config file:

```
# Added by Krypton
Host *
	IdentityAgent ~/.kr/krd-agent.sock
	ProxyCommand /usr/local/bin/krssh %h %p
	IdentityFile ~/.ssh/id_krypton
	IdentityFile ~/.ssh/id_ed25519
	IdentityFile ~/.ssh/id_rsa
	IdentityFile ~/.ssh/id_ecdsa
	IdentityFile ~/.ssh/id_dsa
```

When using SSH with agent forwarding, this works as expected.  I can ssh to my remote server, then ssh elsewhere, and the agent requests come back to my local machine, runs the ProxyCommand, and krssh pings my phone for access.

However, when using `et -f`, I get this when I try to ssh out:

```
$ ssh localhost
isaacs@localhost: Permission denied (publickey).

$ git clone git@github.com:npm/cli
Cloning into 'cli'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

If I add my non-krypton id_rsa.pub to `~/.ssh/authorized_keys` on the server, then  I can `ssh localhost` just fine from within the et session, so it _is_ forwarding agent requests in that case, but it seems like it's not forwarding agent requests in a way that supports IdentityAgent and/or ProxyCommand.

I am not an ssh expert by any means, but it looks like the hard-coded `SSH_AUTH_SOCK` environ lookup [in TerminalClient.cpp](https://github.com/MisterTea/EternalTerminal/blob/master/src/terminal/TerminalClient.cpp#L85-L98) would need to dereference against the SSH config value of IdentityAgent (which can be either the literal string `SSH_AUTH_SOCK` or a path to a socket, potentially containing `~` to mean env.HOME), and use that to get the socket location to send through the connection.  I'm not sure how the `ProxyCommand` option factors into that, though.

I think it would be good for et to work with krypton, since it's a much more secure way to manage private keys.

isaacs:
Fixed by https://github.com/MisterTea/EternalTerminal/pull/249

PR #249, opened 2019-11-26
Add CLI flag for SSH socket to forward and config file parsing

jgoguen:
Adds a CLI flag to specify the SSH agent socket to forward. Although the
socket held in `SSH_AUTH_SOCK` is the default, multiple agents may run
and this allows forwarding a specific agent.

Both SSH agent forwarding and a specific SSH agent socket to use may be
specified in `ssh_config` or `~/.ssh/config`, using the `ForwardAgent`
and `IdentityAgent` parameters. EternalTerminal now knows how to parse
those options and will use them to forward an SSH agent socket and
choose the correct socket based on the values set in the config file i
CLI flags are not given.

Verified by connecting to ET using no CLI flags, only
`--forward-ssh-agent`, only `--ssh-socket`, and both CLI flags. Sockets
were correctly forwarded, with `--ssh-socket` taking priority over a
value set in `ssh_config`.

PR #263, opened 2019-12-04
Add support for IdentityAgent and ForwardAgent in ssh config files

isaacs:
This commit turns on SSH agent forwarding `-f` by default when `ForwardAgent yes` is found in an ssh config file, and uses the `IdentityAgent` if specified rather than only looking at the `SSH_AUTH_SOCK` variable, making it possible to use et with krypton.

Fix https://github.com/MisterTea/EternalTerminal/issues/262

### DIFF
--- a/src/terminal/ParseConfigFile.hpp
+++ b/src/terminal/ParseConfigFile.hpp
@@ -68,6 +68,7 @@ enum ssh_config_opcode_e {
   SOC_STRICTHOSTKEYCHECK,
   SOC_KNOWNHOSTS,
   SOC_PROXYCOMMAND,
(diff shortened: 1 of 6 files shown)


isaacs:
I'm not sure the best way to go about writing tests for this, but it feels like it should probably have some?

isaacs:
Oh, looks like this is a nearly-identical dupe of #249.

Feel free to merge that one instead of this one, @jgoguen's C++ looks a little cleaner than mine :)

MisterTea:
Hey, thanks for taking a crack at this.  #249 should go in soon.

PR #264, opened 2019-12-05
test ssh flag with fix

MisterTea:


MisterTea:
(Not submitting this.  The changes will go into #249 )

PR #269, opened 2019-12-26
Add execinfo as option

ndowens:
libexecinfo is not native to MUSL libc, and this
allows one to either enable/disable the need
to have libexecinfo on their system.

I will be submitting a PR for the external_imported/easyloggingcpp to add support for optional execinfo as well.

Signed-off-by: Nathan Owens <ndowens04@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,11 @@ endif()
 
 option(CODE_COVERAGE "Enable code coverage" OFF)
 option(DISABLE_CRASH_LOG "Disable installing easylogging crash handler" OFF)
+option(ENABLE_EXECINFO "Enable execinfo support" OFF)
+
+if(ENABLE_EXECINFO)
+	SET(CMAKE_LINKER_FLAGS "${CMAKE_LINKER_FLAGS} -lexecinfo")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -12,6 +12,11 @@ endif()
 
 option(CODE_COVERAGE "Enable code coverage" OFF)
 option(DISABLE_CRASH_LOG "Disable installing easylogging crash handler" OFF)
+option(ENABLE_EXECINFO "Enable execinfo support" OFF)
+
+if(ENABLE_EXECINFO)
+	SET(CMAKE_LINKER_FLAGS "${CMAKE_LINKER_FLAGS} -lexecinfo")
+endif()
```
MisterTea: Instead of doing this, please change the logic below that adds execinfo to CORE_LIBRARIES for FreeBSD and NetBSD

ndowens:
Looks like it is already that way

CMakeLists.txt
ndowens: Since after relooking at this, the issue is mostly the easylogging library from the other vender, so closing.

Issue #245, opened 2019-11-05
Fails to install on Debian testing


haxwithaxe:
Using Debian sid (testing) and running the install instructions for Debian verbatim produces this error
```
# apt install et
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 et : Depends: libprotobuf-lite10 but it is not installable
      Depends: libsodium18 (>= 0.6.0) but it is not installable
E: Unable to correct problems, you have held broken packages.
```
Changing the Debian release to buster, sid, or testing results in a 404 error when running `apt update`
It would be a good idea to have packages for the current stable release (buster), and nice to have them for testing (sid).

arthurk:
Same problem here. The apt repo is not working `https://mistertea.github.io/debian-et/debian-source/` it shows 404 error

yix:
Debian Buster is crying...

MisterTea:
@haxwithaxe Try now, I rewrote the debian deployment script.

oezh:
Still not working:

# apt install et
Reading package lists... Done
Building dependency tree
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 et : Depends: libprotobuf-lite10 but it is not installable
      Depends: libsodium18 (>= 0.6.0) but it is not installable
E: Unable to correct problems, you have held broken packages.

foriequal0:
The debian source in the homepage should be changed to
```
deb https://github.com/MisterTea/debian-et/raw/master/debian-source/ buster main
```
`github.io` doesn't seem to properly serve file when it is stored in the Github LFS.

PR #276, opened 2020-02-04
Update Debian installation instruction

foriequal0:
`github.io` doesn't seem to properly serve files stored in Github LFS. It only serves LFS metadata, not the whole file.
Also Debian stretch source doesn't work with Debian buster, which is the most recent stable release.
Resolves: https://github.com/MisterTea/EternalTerminal/issues/245


MisterTea:
Thanks!

PR #280, opened 2020-02-11
Create passkey on server side

MisterTea:


src/terminal/TerminalMain.cpp
```diff
@@ -103,8 +103,18 @@ int main(int argc, char** argv) {
...
+        FATAL_FAIL(setenv("TERM", tokens[1].c_str(), 1));
+      } else if (tokens.size() == 1) {
+        // Generate a new idpasskey for this session
+        string passkey = genRandomAlphaNum(32);
+        string id = genRandomAlphaNum(16);
+        idpasskey = id + string("/") + passkey;
+        FATAL_FAIL(setenv("TERM", tokens[0].c_str(), 1));
+      } else {
+        LOG(FATAL) << "Invalid number of tokens: " << tokens.size();
+      }
```
ailzhang: Btw there're two more `cout` with passkey in this file that we need to remove. (links were in workchat ;)

PR #287, opened 2020-02-25
support 'Match host' in SSH config files

johansglock:
Currently 'Match host' is ignored in SSH options parsing, this can lead
to the situation where if the next line is (for example) 'User foo',
that this option gets applied globally.

Fix this by treating any line that has the form of 'Match <tokens>' as
'[tokens]'. This way 'Match host <options>' gets treated as 'host [options]'.


MisterTea:
Thanks!

PR #306, opened 2020-04-18
Fix errno being overwritten by vlog calls in UnixSocketHandler::accep…

chipturner:
…t which was causing fatals in the branch at the bottom of the function.


MisterTea:
Thanks!!

MisterTea:
I found other places where this was an issue and created #307 to fix all of them

PR #312, opened 2020-05-13
README: Use "now" option instead of start command

axxx007xxxz:
It doesn't really change anything, just less characters to type. lol


MisterTea:
Thanks!

PR #320, opened 2020-06-09
Add installation instructions for CentOS 8

PegasisForever:
You currently can't build et on CentOS 8 due to the lack of `protobuf-lite-devel` package, I added a workaround using packages from The Fedora repository in README.md.


PegasisForever:
Wait, et is already in the epel so no need to grab it from the Fedora repository.

PavelSosin-320:
On Windows 2004 WSL 2 CentOS8 installation works in 3 steps:
1. Download rpm from the repository using a browser to user' download
2. Open CentOS in WT -> PWD = C:\Users\user
3. Run dnf install ./downloads/EternalTerminal-6.0.7-lp152.1.3.x86_64.rpm

PegasisForever:
Does the following work?

```
sudo dnf install epel-release
sudo dnf install et
```

It would be best if you can install it from some kind of repository so you don't need manually update it.

MisterTea:
@PegasisForever I'm not sure I understand this PR?  You have instructions but also say it won't work?

PegasisForever:
Just want to add install instructions for CentOS 8. (`sudo dnf install epel-release && sudo dnf install et`)
Because otherwise people will follow the "[Other Linux](https://github.com/MisterTea/EternalTerminal#other-linux)" and try to compile it (which will not work)

MisterTea:
Let me add a cmake option to link to full protobuf and fix centos

MisterTea:
Hey @PegasisForever can you update this PR to point people to the dockerfile that builds ET correctly on centos 8?  See https://github.com/MisterTea/EternalTerminal/pull/334 for details

PegasisForever:
@MisterTea I think you should add `--recurse-submodules` to [deployment/centos.Dockerfile#L27](https://github.com/MisterTea/EternalTerminal/blob/906fbca145deebe77ea4d6140f11ff89b7002488/deployment/centos.Dockerfile#L27)? Or git won't clone submodules requred to compile.

MisterTea:
The release branch does not have submodules. 

The reason I copy a private key into the docker container is because the docker image is for deploying new releases of ET to various package managers. 

I could use https instead of ssh or we could have a separate docker container for people wanting to use ET, but it seems silly since we have the packages.

PegasisForever:
Thanks for the explaination! I updated the readme and edited the docker files to use https.

One check is failing but I don't think it's related to my PR because I didn't change any source code?

MisterTea:
@PegasisForever please rebase or merge in master.  I fixed that flaky test.

PegasisForever:
Merged to master

MisterTea:
Thanks!

PR #329, opened 2020-07-10
Fix debian installation instructions

coderkd10:
Follow redirects with curl using `-L` option while adding debian gpg key.

without `-L`  gpg errors out "no valid OpenPGP data found"
with `-L` installation works!


MisterTea:
Thanks!

PR #345, opened 2020-08-08
Do not depend on dev packages

SuperSandro2000:


MisterTea:
Depending on dev packages does create some bloat, but the alternative is to depend on a version that changes across distributions and is a hassle to maintain.

PR #374, opened 2021-01-04
Add Android support

buttaface:
Works both completely locally in Termux without root, and when connecting to a linux server. Only remaining issue is that `etserver --help` doesn't work: I think that's because `cxxopts` is incompatible with LLVM libc++, as this patch runs fine on linux x86_64 with GNU libstdc++. I will look into that and patch that last bug later.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,7 +118,9 @@ add_custom_target(
   ${ETERMINAL_SRCS} ${ETERMINAL_HDRS}
   )
 
-IF(FREEBSD)
+IF(ANDROID)
+  set(CORE_LIBRARIES util)
+ELSEIF(FREEBSD)
   set(CORE_LIBRARIES util execinfo)
 ELSEIF(NETBSD)
   set(CORE_LIBRARIES util resolv execinfo)
@@ -133,7 +135,11 @@ ENDIF()
 MACRO(DECORATE_TARGET TARGET_NAME)
   add_sanitizers(${TARGET_NAME})
   set_target_properties(${TARGET_NAME} PROPERTIES COTIRE_CXX_PREFIX_HEADER_INIT "src/base/Headers.hpp")
-  cotire(${TARGET_NAME})
+  IF(CMAKE_CROSSCOMPILING)
+    # Doesn't work when cross-compiling
+  ELSE()
+    cotire(${TARGET_NAME})
+  ENDIF()
 ENDMACRO()
 
 include_directories(
@@ -142,7 +148,6 @@ include_directories(
   external/Catch2/single_include
   external/cxxopts/include
   external/msgpack-c/include
-  external/UniversalStacktrace/ust
   external/sentry-native/include
   src/base
   src/terminal
@@ -156,6 +161,10 @@ include_directories(
   ${SELINUX_INCLUDE_DIR}
   ${UTEMPTER_INCLUDE_DIR}
   )
+
+IF(NOT ANDROID)
+  include_directories( external/UniversalStacktrace/ust )
+ENDIF()
 
 add_library(
   et-lib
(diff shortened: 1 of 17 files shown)


CMakeLists.txt
```diff
@@ -157,6 +162,10 @@ include_directories(
   ${UTEMPTER_INCLUDE_DIR}
   )
 
+IF(NOT ANDROID)
+  include_directories( external/UniversalStacktrace/ust )
```
buttaface, line 166: No libexecinfo on Android.

src/base/Headers.hpp
```diff
@@ -288,7 +296,7 @@ inline string GetTempDirectory() {
   std::wstring_convert<std::codecvt_utf8_utf16<wchar_t> > converter;
   std::string tmpDir = converter.to_bytes(wstring(buf, retval));
 #else
-  string tmpDir = "/tmp";
+  string tmpDir = _PATH_TMP;
```
buttaface, line 299: Adds a trailing slash, the Windows one may need to be checked for that.

src/base/TcpSocketHandler.cpp
```diff
@@ -12,7 +12,7 @@ int TcpSocketHandler::connect(const SocketEndpoint &endpoint) {
   memset(&hints, 0, sizeof(addrinfo));
   hints.ai_family = AF_UNSPEC;
   hints.ai_socktype = SOCK_STREAM;
-#if __NetBSD__
+#if defined(__NetBSD__) || defined(__ANDROID__)
   hints.ai_flags = (AI_CANONNAME | AI_ADDRCONFIG);
```
buttaface, line 16: [As noted on SO, Android doesn't support these flags](https://stackoverflow.com/questions/39674147/getaddrinfo-on-android-returning-error-eai-badflags) just like BSD, which is where its source comes from.

src/terminal/ParseConfigFile.hpp
```diff
@@ -34,6 +34,8 @@
 #include <sys/time.h>
 #endif /* HAVE_SYS_TIME_H */
 
+#include "Headers.hpp"
```
buttaface, line 37: This was missing when cross-compiling to Android, but is reproducible on linux too if cotire is turned off. I guess precompiling the headers pulls in other headers that weren't included originally.

CMakeLists.txt
MisterTea: ust uses libunwind now, can you try latest master and tell me what error you get if any?

src/htm/HtmClientMain.cpp
```diff
@@ -64,9 +64,9 @@ int main(int argc, char** argv) {
   el::Loggers::setVerboseLevel(3);
   // default max log file size is 20MB for etserver
   string maxlogsize = "20971520";
-  LogHandler::setupLogFile(&defaultConf, "/tmp/htm.log", maxlogsize);
+  LogHandler::setupLogFile(&defaultConf, GetTempDirectory() + "htm.log", maxlogsize);
```
MisterTea, line 67: Thanks for cleaning this up!

CMakeLists.txt
buttaface: Doesn't seem to work, as Android uses the LLVM libunwind but UniversalStacktrace seems to expect the nongnu one? I get errors about missing symbols like `UNW_REG_IP` when compiling, after changing `external/UniversalStacktrace/ust/ust.hpp` to include `unwind.h` instead of `libunwind.h`.

buttaface:
Updated to replace `/tmp` in the tests too and built trunk natively on Android for the first time, passes all tests natively on Android.

buttaface:
I found the cause of `etserver --help` choking: it's because one of the help strings was a certain length, worked around in termux/termux-packages@79de94f. Probably also related to libc++, still unsure why it happens as I haven't dug that deep.

PR #377, opened 2021-01-10
Run EternalTerminal systemd service with Restart=on-failure

koppchen:
With Restart=on-failure, the service will be restarted when the process
exits with a non-zero exit code, is terminated by a signal, when an
operation (such as service reload) times out, and when the configured
watchdog timeout is triggered.

Setting this to on-failure is the recommended choice for long-running
services, in order to increase reliability by attempting automatic
recovery from errors.

### DIFF
--- a/systemctl/et.service
+++ b/systemctl/et.service
@@ -4,6 +4,7 @@ After=syslog.target network.target
 
 [Service]
 Type=simple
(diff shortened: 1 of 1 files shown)


MisterTea:
Thanks!

Issue #72, opened 2017-11-13
Allow to pass options to SSH


DmitryKoterov:
In the previous et versions, there was a `-s` switch to pass additional options to ssh. E.g. it allowed to pass `-o PasswordAuthentication=no` or something else. (I understand it was mostly used for jumphosts and similar stuff, but its usage was still wider than that.)

In newer versions, this switch is removed, and there is no way to pass options to the initial ssh process except putting them to .ssh/config.

What do you think about adding this feature back? Or allowing `et` to accept `-o aaa=bbb -o ccc=ddd` the same way as ssh does?

ailzhang:
Hi Dmitry,
We think using .ssh/config is a cleaner way, is there any reason that sshconfig is not working for your case?

DmitryKoterov:
Yes, the case is when et is used as a part of some other tool (e.g. to setup a tunnel and then do something else with this tunnel repeatedly - e.g. run Unison over it); this tool is self-contained (and even has an installer) and doesn't want to make changes to user's .ssh/config file.

pavel64-sutyrin:
Same issue here. `/.ssh/config` seems to be not picked up (I am interested in `ForwardAgent`). Pure `ssh` works, `et` does not work, with all the same actions.

MisterTea:
@pavel64-sutyrin We support agent forwarding now

MisterTea:
@DmitryKoterov I'm fine with adding a switch for custom switch for ssh options.  You want to add it to TerminalClientMain.cpp and then pass the options to SshSetuphandler

gklanderman:
close as resolved?

PR #390, opened 2021-02-09
Add --ssh-option argument for passing arguments to ssh

wez:
We'd like to be able to pass arbitrary ssh `-o` config options
down to the underlying ssh command.

Following the suggestions in the last comment on:
https://github.com/MisterTea/EternalTerminal/issues/72

this commit adds a vector of string arguments that can be passed down.

As part of this, tidy up construction of the ssh args list to remove
some redundant string -> c_string -> string conversion and clarify the
control flow a bit.

Test Plan:

I tried running this, with an option picked at random to have
no real effect other than showing up in the argument list:

```
$ ./et --ssh-option RequestTTY=yes SOME.HOST
```

then inspecting the process listing I can see that it made it
into the ssh command that gets run:

```
ssh USER@SHOME.HOST -oRequestTTY=yes echo 'XXXXXXX_xterm-256color\012' |  etterminal --verbose=0
```

### DIFF
--- a/src/terminal/SshSetupHandler.cpp
+++ b/src/terminal/SshSetupHandler.cpp
@@ -24,7 +24,8 @@ string SshSetupHandler::SetupSsh(const string& user, const string& host,
                                  const string& host_alias, int port,
                                  const string& jumphost, int jport, bool kill,
                                  int vlevel, const string& cmd_prefix,
-                                 const string& serverFifo) {
+                                 const string& serverFifo,
+                                 const std::vector<std::string>& ssh_options) {
   string clientTerm("xterm-256color");
   auto envString = getenv("TERM");
   if (envString != NULL) {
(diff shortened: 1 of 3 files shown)


src/terminal/SshSetupHandler.cpp
```diff
@@ -50,17 +51,24 @@ string SshSetupHandler::SetupSsh(const string& user, const string& host,
...
+    ssh_args = {
+      "-J",
+      SSH_USER_PREFIX + jumphost,
+    };
+  }
+
+  ssh_args.push_back(SSH_USER_PREFIX + host_alias);
+
+  for (auto& arg : ssh_options) {
+    ssh_args.push_back("-o" + arg);
```
MisterTea, line 65: You need a space after the -o, no?

MisterTea (changes requested):

src/terminal/SshSetupHandler.cpp
wez: No space required!

wez:
The tsan failure appears to be a pre-existing issue that might be resolved by copying this mutex acquisition:
https://github.com/MisterTea/EternalTerminal/blob/master/src/base/SocketHandler.hpp#L44
so that it also happens before the `readAll` call here:
https://github.com/MisterTea/EternalTerminal/blob/master/src/base/SocketHandler.hpp#L24

MisterTea:
> The tsan failure appears to be a pre-existing issue that might be resolved by copying this mutex acquisition:
> https://github.com/MisterTea/EternalTerminal/blob/master/src/base/SocketHandler.hpp#L44
> so that it also happens before the `readAll` call here:
> https://github.com/MisterTea/EternalTerminal/blob/master/src/base/SocketHandler.hpp#L24

Yeah it's a bug in libunwind but I'm using a mutex to patch it.  Please see my comment about the missing space

wez:
> Please see my comment about the missing space

I responded inline above; no space is necessary!

MisterTea:
Thanks!

Issue #382, opened 2021-01-28
Ask client to confirm telemetry on launch


SuperSandro2000:
Will there be an option to disable telemetry entirely?

MisterTea:
There's a telemetry option in et.cfg to disable it server-side and there's the --telemetry option to disable it on the client.

SuperSandro2000:
Did you think about disabling it by default? Especially in the open source community it is considered not nice and sometimes even rude to enable telemetry by default.

MisterTea:
Can you elaborate on that?

I'm collecting crash logs using crashpad, and also collecting LOG(ERROR) and LOG(FATAL).  The only thing I'm using the logs for is to make ET better.

SuperSandro2000:
You are sending the data to sentry which most likely collects personal data which probably requires a privacy policy. So far the readme does not contain any information about this and trying to hide such things looks always very suspicious even you just try to get more insights into crashes. If you would be open about it, would make it easy to disable and opt-in rather than opt-out it would be less of a problem for most people.

MisterTea:
Sentry only collects the data I send it (and possibly the IP address of the sender)

MisterTea:
@SuperSandro2000 Please take a look at #395 and give me your thoughts.

PR #395, opened 2021-02-16
Alert user about telemetry and give instructions to remove.  Remove IP address from logging. #382

MisterTea:
…P address from logging.  #382


SuperSandro2000:
Nice. I would still like it to be opt in but better than no opt out.

PR #421, opened 2021-04-02
Fix occasional Host missmatches

razum2um:
Backported libssh commit 1ba6ef689fe7f702d9921271dd422cae885de788

I'm having quite big `.ssh/config` and a very mysterious issue like this.
Note, hosts are different, should not match, x.x.x.x is _not in the config at all_
```
./et --logtostdout x.x.x.x
Parsed ssh config file, connecting to y.y.y.y
```

On my x64-osx(11.2.2) it happens at least in 10% with a single-host config (not sure what's special in it) and 100% in my working config with >50 hosts, but clear reproduce steps are hard. However, on my laptop this is reproducible with both clang-12.0.0 and gcc-10.2.0

From my understanding it happens like this:

- this [SOC_HOST parsing for loop](https://github.com/MisterTea/EternalTerminal/blob/b11784967704dcd86bbbe2fb930595f0c9404cf8/src/terminal/ParseConfigFile.hpp#L1227L1228) consumes actual hostname c_str (after `Host` c_str just consumed from `*str`)
- the line causing the issue is `*str = c + 1;` which sets parseable pointer _behind_ hostname
- then it depends on memory layout - sometimes this is not a null byte
- the most mysterious part: this memory offset contains exactly string with `x.x.x.x`
- the loop kicks in one more time, in this last loop `parsing = 1` happens
- then `SOC_HOSTNAME` switch-part happens (nethertheless we're _on the wrong hostname line_ now)
- so it consumes wrong `HostName` from this section

Commit I refer here can by seen on github: https://github.com/simonsj/libssh/commit/1ba6ef689fe7f702d9921271dd422cae885de788

### DIFF
--- a/src/terminal/ParseConfigFile.hpp
+++ b/src/terminal/ParseConfigFile.hpp
@@ -1076,20 +1076,49 @@ out:
 
 static char *ssh_config_get_token(char **str) {
   char *c;
-  char *r;
+  bool had_equal = false;
+  char *r = NULL;
 
-  c = ssh_config_get_cmd(str);
-
-  for (r = c; *c; c++) {
-    if (isblank(*c) || *c == '=') {
-      *c = '\0';
-      goto out;
-    }
+  /* Ignore leading spaces */
+  for (c = *str; *c; c++) {
+      if (! isblank(*c)) {
+          break;
+      }
   }
 
-out:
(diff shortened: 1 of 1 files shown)


MisterTea:
Thanks!

PR #424, opened 2021-04-18
Execute shell with -l rather than --login for dash compatibility

lnikkila:
I use dash as my login shell and noticed that etterminal launches SHELL with --login, which dash doesn't support.

Since -l seems to be supported by the major shells like bash and zsh, I'm wondering whether this could be an appropriate change to support dash as well.

Thanks!


MisterTea:
Thanks for the PR!

Issue #366, opened 2020-12-24
et reconnection when using iterm tmux mode is very slow when waking from sleep


christopherhesse:
This app looks pretty great, but I'm having an issue which makes it hard to use on Mac OS X. 

Specifically, I use the iterm tmux control mode over ssh to connect to a remote server.  With ssh, of course, the connection dies when my laptop goes to sleep and I have to reconnect manually on wake.  With et, I thought the connection was dying permanently on wake, but actually it's waiting 60 seconds before reconnecting.

I enabled et logging, but the log appears to be empty.  With iterm tmux logging, I see my keystrokes being logged, but no response from the server until 60 seconds have elapsed.

This makes me think there is some timeout that is being triggered.  Disconnecting from my VPN and the reconnecting also takes like 30 for et to reconnect.

Any idea why the connection dying is not being detected?  mosh + tmux works fine in both situations, generally taking only a second or two.

MisterTea:
ET should take about 10 seconds to reconnect (beyond the time it takes to reconnect to the internet).  You could turn verbose logging on with the " -v 9" flag and that will provide a lot more information.  ET does not cache your dns lookup, maybe your DNS server is really slow?

christopherhesse:
Thanks for the quick response!  I did use verbose logging but the log file was empty, does `--verbose=9` log to `/tmp/etclient_stderr*`?  Would I be able to see the slow dns server with `nslookup`?

rodolf0:
Had a similar issue. After turning wifi off and on again, or waking from sleep, re-connection would take forever. Turns out I was connecting to a cname (alias) instead of the underlying host name and the lookup would timeout. Connecting directly to the hostname solved this for me.

rodolf0:
I'm actually wrong. Seems this is unrelated to cname. Seems to be hanging on a long timeout while calling connect. Here's the pointer to the code: https://github.com/MisterTea/EternalTerminal/blob/master/src/base/TcpSocketHandler.cpp#L55-L64

Logs:

```
[INFO 2021-04-18 16:32:34,383 Reconnect TcpSocketHandler.cpp:59] Error connecting with devvm2958.theserver.com: 101 Network is unreachable
[INFO 2021-04-18 16:32:34,383 Reconnect TcpSocketHandler.cpp:152] No host found
[INFO 2021-04-18 16:32:35,384 Reconnect TcpSocketHandler.cpp:59] Error connecting with devvm2958.theserver.com: 101 Network is unreachable
[...]

PR #426, opened 2021-04-19
Mark socket as non-blocking before calling connect (#366)

rodolf0:
Reconnection is sometimes slow because 'connect' cannot establish
connection immediately and blocks until underlying timeout.

This change marks the socket non-blocking making use of the fact that
we're later calling select with a low timeout allowing fast failure.

### DIFF
--- a/src/base/TcpSocketHandler.cpp
+++ b/src/base/TcpSocketHandler.cpp
@@ -52,6 +52,7 @@ int TcpSocketHandler::connect(const SocketEndpoint &endpoint) {
       continue;
     }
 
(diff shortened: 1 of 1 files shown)


rodolf0:
@MisterTea  this addresses https://github.com/MisterTea/EternalTerminal/issues/366

MisterTea:
Thanks for this PR!

I dug into it and this ended up being a much bigger change to support all OS's so I made a new PR here: https://github.com/MisterTea/EternalTerminal/pull/435

Issue #423, opened 2021-04-17
server crash: arm64 ubuntu18.04 crash at client connect


yurenchen000:
et arm64 ubuntu crash
=====================

## 一. version

// server, use same et conn localhost also crash
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.2 LTS"

$ apt list et
Listing... Done
et/bionic,now 6.1.7-bionic2 arm64 [installed]

$ uname -a
Linux HostName 4.4.194 #6 SMP Fri Dec 18 17:51:03 CST 2020 aarch64 aarch64 aarch64 GNU/Linux

// client, connect other x64 et server ok
$ apt list et
et/focal,now 6.1.7-focal2 amd64 [installed]

$ uname -a
Linux chen-x1c 5.4.0-62-generic #70-Ubuntu SMP Tue Jan 12 12:45:47 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

## 二. err

run `et localhost` got failed, then find server crashed.

```
systemctl status et
● et.service - Eternal Terminal
   Loaded: loaded (/lib/systemd/system/et.service; enabled; vendor preset: enabled)
   Active: failed (Result: signal) since Sun 2021-04-18 04:49:20 CST; 5s ago
  Process: 12053 ExecStart=/usr/bin/etserver --cfgfile=/etc/et.cfg (code=killed, signal=SEGV)
 Main PID: 12053 (code=killed, signal=SEGV)

Apr 18 04:49:14 HostName systemd[1]: Started Eternal Terminal.
Apr 18 04:49:20 HostName systemd[1]: et.service: Main process exited, code=killed, status=11/SEGV
Apr 18 04:49:20 HostName systemd[1]: et.service: Failed with result 'signal'.
```

### server

```
/usr/bin/etserver --cfgfile=/etc/et.cfg --telemetry --logtostdout --verbose 99 
[INFO 2021-04-18 04:58:55,409 etserver-main TerminalServerMain.cpp:168] In child, about to start server.
[INFO 2021-04-18 04:58:55,411 etserver-main TcpSocketHandler.cpp:231] Listening on 0.0.0.0:2022/2/1/6
[...]

PR #431, opened 2021-04-25
fix: expand Arch List with aarch64 

Jacke:
that is the official name for arm64-linux arch on the latest Ubuntu distr.
This will fix the building from source to this issue: https://github.com/MisterTea/EternalTerminal/issues/423


Jacke:
Please do not merge this PR

I got the same error 
```
❯ ./etserver -v10
[1]    8007 segmentation fault (core dumped)  ./etserver -v10
```

MisterTea:
etserver needs to be run as root

Jacke:
@MisterTea yep, if I try to run etserver as root it works. Thanks!

PR #442, opened 2021-05-25
Add --macserver option

MisterTea:


tumd:
When using Homebrew the [default prefix](https://docs.brew.sh/FAQ#why-should-i-install-homebrew-in-the-default-location) unfortunately varies, and is not always `/usr/local`. For whatever reason it's `/opt/homebrew` on Apple Silicon Macs.

PR #462, opened 2021-09-13
Telemetry for Mojave

dgentry:
This makes Telemetry work on Mojave:

1.  Install boost with `brew install boost`
2.  Cmake finds Boost
3. `TelemetryService.hpp` sets `fs` to a filesystem namespace that it can find (`boost`, `std`, or `std::experimental`)
4. `TelemetryService.cpp' uses `fs::` instead of `filesystem::`

I imagine we'll need to simultaneously change homebrew-et/et.rb to include dependencies:

```
  depends_on "boost"
  depends_on "ninja"
```

(`ninja` is just missing now; `boost` is of course a new dependency.). I don't know brew well enough to know if it's possible to only add boost as a dependency for Mojave systems and not for Catalina onward.

Compiled and tested (`et-test`) on x86 Mojave, M1 Big Sur, and PocketBeagle Debian Buster (kernel 4.19.94-ti-r64).

I would appreciate any feedback you have on this PR!

### DIFF
--- a/src/terminal/TelemetryService.hpp
+++ b/src/terminal/TelemetryService.hpp
@@ -4,18 +4,16 @@
 
 #ifndef NO_TELEMETRY
 #if __has_include(<boost/filesystem.hpp>)
-#pragma message "Using boost::filesystem"
 #include <boost/filesystem.hpp>
 namespace fs = boost::filesystem;
 #elif __has_include(<filesystem>)
 #include <filesystem>
 namespace fs = filesystem;
 #elif __has_include(<experimental/filesystem>)
-#pragma message "Using std::experimental::filesystem"
 #include <experimental/filesystem>
 namespace fs = std::experimental;
 #else
-#pragma message "No filesystem library found."
+#pragma message "No filesystem library found (which Telemetry needs)."
 #endif
 #endif
 
--- a/src/terminal/TelemetryService.hpp
+++ b/src/terminal/TelemetryService.hpp
@@ -2,6 +2,23 @@
 
 #include "Headers.hpp"
 
+#ifndef NO_TELEMETRY
+#if __has_include(<boost/filesystem.hpp>)
+#pragma message "Using boost::filesystem"
+#include <boost/filesystem.hpp>
+namespace fs = boost::filesystem;
+#elif __has_include(<filesystem>)
+#include <filesystem>
+namespace fs = filesystem;
+#elif __has_include(<experimental/filesystem>)
+#pragma message "Using std::experimental::filesystem"
+#include <experimental/filesystem>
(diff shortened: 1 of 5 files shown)


src/terminal/TelemetryService.hpp
```diff
@@ -2,6 +2,21 @@
 
 #include "Headers.hpp"
 
+#ifndef NO_TELEMETRY
```
MisterTea, line 5: Move this into Headers.hpp so it's pre-compiled and in case other files want to use filesystem functions.

MisterTea: also I think we can get rid of the NO_TELEMETRY wrapper and allow filesystem functions without telemetry.

MisterTea (changes requested):
Please address inline comments, thanks!

src/terminal/TelemetryService.hpp
dgentry: Will do.  

The original code in `Header.hpp` was `using namespace <x>;` and just using `boost` caused ambiguities in unrelated .cpp code, so moving the whole condition to `Telemetry.hpp` avoided that.  But I also realized that defining an unambiguous `fs` namespace and using that was a better approach anyway, so yeah, it'll be totally safe to move it back to `Header.`

I'm building and testing it now on a few weird old systems, and I'll push a commit when they finish.  The PocketBeagle takes about an hour and a half to build.  (That's not counting the time to build a CMake version new enough to work.). The M1 Mac takes about a minute.

dgentry: What I pushed works on modern Debian/Ubuntu, but fails on Raspberry Pi and Debian buster, so I'm testing better conditionalization now.  I'll push again once I've tested on a few systems.

dgentry: It required a bit of specific conditionalization so it only uses `boost::filesystem` when `std::filesystem` isn't there (but not on Mojave because `std::filesystem` is there but not usable).  Without that, it failed to link on systems with older Boost (latest Raspbian, for example).  But it now passes `et-test` everywhere I built it.

I recommend squashing all these commits, if that's an option.

dgentry:
Discovered that the README instructions to start a LaunchDaemon for m1 Macs only worked if you were already root, so fixed the instructions.  As far as I know, if the CI passes, this PR is ready to go.

MisterTea:
> Hi, should I do anything else to get this PR reviewed again/merged? I wasn't sure if I should mark the conversation above as resolved or leave that for you to do.
> 
> Thanks.

Thanks for bumping this up!  Looks great. Thanks again for the PR.

PR #471, opened 2021-11-04
Add deb packaging instructions and support

pjz:


pjz:
Cmake has decent support for building .debs, why not use it?

MisterTea:
Learn something new every day

pjz:
Given that I only touched the build system, I don't understand how code coverage has decreased.  Anyway, is there something I need to do to make this more acceptable?

PR #479, opened 2021-12-25
add configurable client keep-alive duration

cfal:
hello! thanks for a great tool. this PR (should) make the client keep-alive duration configurable.

i created this because i noticed that reconnection times seemed to be slow sometimes (especially when compared to mosh) and am suffering frequently from lost keys. i realize that UDP is probably better suited for cases where there is frequent switching but many public wifi networks (at least in my locale) has some sort of filtering in place that makes EternalTerminal much more practical to use.

disclaimer: i don't have any actual data that this would improve the situation for my use case, because i'm still having issues building on MacOS M1. but this seemed like a fast enough change that i thought i'd make it and get some feedback :)

please take a look, thanks!

### DIFF
--- a/src/terminal/TerminalClient.cpp
+++ b/src/terminal/TerminalClient.cpp
@@ -66,8 +66,9 @@ TerminalClient::TerminalClient(shared_ptr<SocketHandler> _socketHandler,
                                const string& tunnels,
                                const string& reverseTunnels,
                                bool forwardSshAgent,
-                               const string& identityAgent)
-    : console(_console), shuttingDown(false) {
+                               const string& identityAgent,
+                               int _keepaliveDuration)
+  : console(_console), shuttingDown(false), keepaliveDuration(_keepaliveDuration) {
   portForwardHandler = shared_ptr<PortForwardHandler>(
       new PortForwardHandler(_socketHandler, _pipeSocketHandler));
   InitialPayload payload;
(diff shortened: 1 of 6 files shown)


src/terminal/TerminalClientMain.cpp
```diff
@@ -157,6 +160,12 @@ int main(int argc, char** argv) {
     string jumphost =
         result.count("jumphost") ? result["jumphost"].as<string>() : "";
     string host_alias = destinationHost;
+    uint32_t keepaliveDuration = result.count("keepalive") ? result["keepalive"].as<uint32_t>() : MAX_CLIENT_KEEP_ALIVE_DURATION;
+    if (keepaliveDuration < 1 || keepaliveDuration > MAX_CLIENT_KEEP_ALIVE_DURATION) {
```
MisterTea, line 164: Please change this to be a signed integer

src/terminal/TerminalClient.cpp
```diff
@@ -66,8 +66,9 @@ TerminalClient::TerminalClient(shared_ptr<SocketHandler> _socketHandler,
                                const string& tunnels,
                                const string& reverseTunnels,
                                bool forwardSshAgent,
-                               const string& identityAgent)
-    : console(_console), shuttingDown(false) {
+                               const string& identityAgent,
+                               uint32_t _keepaliveDuration)
```
MisterTea, line 70: Please change to be a signed integer

MisterTea (changes requested):

MisterTea:
Thanks for this PR!  Overall, it looks great.  I'd like us to stick with signed integers unless we really need to use unsigned.

cfal:
> Thanks for this PR! Overall, it looks great. I'd like us to stick with signed integers unless we really need to use unsigned.

updated to use ints, thanks!

MisterTea:
Thanks!

PR #488, opened 2022-01-23
Fix reconnects on FreeBSD

dfr:
With etserver v6.1.9, attempts to reconnect cause the etserver process
to crash due to an unexpected EGAIN in UnixSocketHandler::close.

Suppressing this by ignoring EAGAIN allows etserver to keep running
but the reconnect still fails since the call to close causes the
select in TerminalServer::runTerminal to complete with EBADF and
an attempt to read from terminalFd which returns EAGAIN (since it
is non-blocking) and the terminal server session ends.

Handling EGAIN in both places allows the reconnect to succeed and
re-join the terminal server session reliably.

### DIFF
--- a/src/terminal/TerminalServer.cpp
+++ b/src/terminal/TerminalServer.cpp
@@ -274,7 +274,7 @@ void TerminalServer::runTerminal(
           tb.set_buffer(s);
           serverClientState->writePacket(
               Packet(TerminalPacketType::TERMINAL_BUFFER, protoToString(tb)));
-        } else if (rc == 0 || errno != EAGAIN) {
+        } else {
           LOG(INFO) << "Terminal session ended";
           run = false;
           break;
--- a/src/terminal/TerminalServer.cpp
+++ b/src/terminal/TerminalServer.cpp
@@ -274,7 +274,7 @@ void TerminalServer::runTerminal(
           tb.set_buffer(s);
           serverClientState->writePacket(
               Packet(TerminalPacketType::TERMINAL_BUFFER, protoToString(tb)));
-        } else {
+        } else if (rc == 0 || errno != EAGAIN) {
           LOG(INFO) << "Terminal session ended";
           run = false;
           break;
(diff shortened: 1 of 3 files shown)


dfr:
Not sure why the connect_with_jumphost test is failing - my changes shouldn't have affected it and I can't re-run the test to see if its just a bit flaky.

src/terminal/TerminalServer.cpp
```diff
@@ -274,7 +274,7 @@ void TerminalServer::runTerminal(
           tb.set_buffer(s);
           serverClientState->writePacket(
               Packet(TerminalPacketType::TERMINAL_BUFFER, protoToString(tb)));
-        } else {
+        } else if (rc == 0 || errno != EAGAIN) {
```
MisterTea, line 277: If you revert this change, do the tests pass?

dfr:
Note: there is still a fairly rare problem with FreeBSD reconnects which I'm still in the process of debugging so I will update this PR when I figure that out (as well as the CI problems).

dfr:
Looks like the circleci tests are still having problems. How can I run these tests locally?

dfr:
The log shows:

```
#!/bin/bash -eo pipefail
sudo /etc/init.d/ssh start; rm -f ~/.ssh/id_rsa*; ssh-keygen -q -N "" -f ~/.ssh/id_rsa; sudo rm ~/.ssh/authorized_keys; cp ~/.ssh/id_rsa.pub ~/.ssh/authorized_keys; rm -f ~/.ssh/known_hosts; ssh -o "StrictHostKeyChecking no" localhost ls
Starting ssh (via systemctl): ssh.service.
rm: cannot remove '/home/circleci/.ssh/authorized_keys': No such file or directory

Exited with code exit status 1
CircleCI received exit code 1
```

jwmcglynn:
I ran into the same circleci failures, and have a fix in #511.

PR #489, opened 2022-01-26
Update TerminalClient.cpp

kurktchiev:
I am unsure if you had added the `exit` in order to get around some other bug or not, but here is a patch to remove the appended exit to any command passed from `-c`

### DIFF
--- a/src/terminal/TerminalClient.cpp
+++ b/src/terminal/TerminalClient.cpp
@@ -211,7 +211,7 @@ void TerminalClient::run(const string& command) {
   if (command.length()) {
     LOG(INFO) << "Got command: " << command;
     et::TerminalBuffer tb;
-    tb.set_buffer(command + "; exit\n");
+    tb.set_buffer(command + "\n");
 
     connection->writePacket(
         Packet(TerminalPacketType::TERMINAL_BUFFER, protoToString(tb)));


MisterTea:
The default behavior of -c is to mimic running ssh with a command, which runs the command and exits.  I use -c for the purpose. 

Please create a new switch for running the command and not exiting.

kurktchiev:
Well that doesn’t make sense, the dependency on duplicating ssh behavior is not explained or make sense when you already have specific SSH options to enable you to connect “the ssh way”. 

Adding another switch will just make things more confusing.

MisterTea:
I typically run this:

et -x -c "tmux -CC a"

When I'm done with my tmux session, which could take weeks, the et session also terminates.

kurktchiev:
What happens when you close your laptop and move 4 feet and wake it up again?

I do exactly the same as above as well, however the moment I move with my laptop not powered on, instead of coming back to my session I basically have to reconnect, at which point, what’s the point of Et?

MisterTea:
I don't think you are following:

I run "et -x -c foo" and foo finishes when I say it does. 

I can close my laptop, change wifi, etc.

But when I'm done running foo, the et connection also closes.

MisterTea:
I've never used tmuxinator.

Tmux shouldn't detatch just because you lose network connection. Et and
tmux will resume once the connection is reestablished

kurktchiev:
tmuxinator is simply a wrapper around tmux to create a session with X window configs. end of the day it executes tmux -CC

kurktchiev:
and as i said with the switches and way I am connected with the appended `;exit` what happens for me is, as soon as I go to sleep my session executes the `exit` and i have to manually restart the session.

MisterTea:
Let's debug the issue of tmux detaching instead of changing the behavior of -c

megalithic:
Sounds like this fell away; @kurktchiev / @MisterTea -- any more interest in having the ability to execute commands inline of a connection to a server _without_ the `; exit`?

kurktchiev:
it hasnt really fallen away, i am still unable to actually hold a connection open constantly with ET without restarting things post laptop hibernation, seemingly its the `;` thats biting me in the behind.

MisterTea:
@kurktchiev we recently fixed some issues around disconnecting with macos.

PR #509, opened 2022-04-22
Create devcontainer and add .clang-format to simplify development

jwmcglynn:
This PR has two changes:
* Add a devcontainer, which sets up an environment that is sufficient to build with both gcc/clang.  This can be launched in docker or in GitHub Codespaces, for example.  I chose ubuntu-21.04 since it supports running on M1.
* Add a `.clang-format` file so that using format-on-save works and can automatically detect the style.

For the devcontainer, it is mostly unmodified from the automatically-generated boilerplate.  The following small changes were made:
* Switched to a newer version of cmake (specifically `ARG REINSTALL_CMAKE_VERSION_FROM_SOURCE="3.23.1"`)
* Installed the dependency packages based on `README.md`, with some small update to use `libcurl4-openssl-dev` (this is noted with a comment).
* Run `git submodule update --init --recursive` after launching to initialize submodules.
* Add a command history volume, based on https://code.visualstudio.com/remote/advancedcontainers/persist-bash-history .  Note that there's also a bashrc change which I did not include in this change, since I like to do this in dotfiles instead.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,3 @@
+---
+BasedOnStyle: Google
+---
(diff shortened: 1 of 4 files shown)


MisterTea:
Please rebase or merge master back so we can rerun ci

jwmcglynn:
I think the "ci/circleci: connect_with_jumphost" failure is some sort of flakiness with that test, since this PR should have no functional changes.

PR #510, opened 2022-04-23
Add simple fuzzers for etserver sockets

jwmcglynn:
etserver has two interfaces, the "server" and "router". Add fuzzers for
each of these, and fix crashes from invalid input to the router
(internal) interface.

The fuzz build currently requires libFuzzer and clang, and can be
invoked with:
```
export CC=/usr/bin/clang
export CXX=/usr/bin/clang++

mkdir build && cd build
cmake -DSANITIZE_ADDRESS=ON -DFUZZING=ON ../
make -j`nproc`
```

To run the fuzzers:
```
mkdir router-corpus
cd router-corpus
../TerminalServerRouterFuzzer -jobs=10 .

mkdir server-corpus
cd server-corpus
../TerminalServerFuzzer -jobs=10 .
```

These will keep running unless they hit a crash.

Note that easyloggingpp still overrides the signal handlers, so doing a
Ctrl-C will trigger a spurious crash report, but these can be safely
ignored.

The integration can still be improved, and this can be integrated with
CI and infra like OSSFuzz at a later point, but this change focuses on
bootstrapping the fuzzers.

I included the code fix in this change, since that was blocking running
the fuzzer for longer periods of time.  The failure is an STFATAL that
gets hit in UserTerminalRouter, which I changed to a graceful failure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,6 +157,7 @@ if(POLICY CMP0058)
 endif()
 
 option(CODE_COVERAGE "Enable code coverage" OFF)
+option(FUZZING "Enable builds for fuzz testing" OFF)
 option(DISABLE_CRASH_LOG "Disable installing easylogging crash handler" OFF)
 
 add_definitions(-DET_VERSION="${PROJECT_VERSION}")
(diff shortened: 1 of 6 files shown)


jwmcglynn:
I think CI failed due to https://www.githubstatus.com/incidents/b40k7ckrs7sp, so I kicked it off again.

PR #514, opened 2022-04-29
Add additional checks to connection path and relocate router fifo

jwmcglynn:
Add additional verification when connecting terminals to etserver:
* Reject more than one terminal connecting per launch.
* Ensure that the launched terminal's passkey matches on connect.
* Fix race condition if more than one terminal launches.

To protect the router fifo path, move it to either /var/run, $XDG_RUNTIME_DIR (with a fallback to $HOME/.local/share), depending on if the etserver process is running as root.

If the $HOME/.local/share directory does not exist, it will be created. Note that as a side effect of this change, if etserver is running as a non-root user it may only be connected by the same user, since it now places the fifo in the user-specific $XDG_RUNTIME_DIR or $HOME directory.

For connecting to the pipe, update etterminal to probe to detect the pipe to use, first trying the root location and then falling back to the user-specific location.

Note that this is *not* a breaking change. Since this detects appropriate location, and since both affected binaries are on the server side (etserver and etterminal), existing clients can connect to the server as long as they did not override --serverfifo back to its original location.

Unittests have been added for both the terminal connection path and server fifo fallback logic, but the server fifo path can also be manually verified by:
- Launching etserver as root, and verify connecting.
- Launching etserver as a non-root user, and verify connecting. Also verify that the pipe is located in either $XDG_RUNTIME_DIR/etserver, or $HOME/.local/share/etserver if $XDG_RUNTIME_DIR is not set.
- Rename $HOME/.local/share and verify it gets recreated. (It is convenient to do this on the devcontainer)
- Rename $HOME/.local and verify it gets recreated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -399,6 +399,8 @@ add_library(
   src/terminal/UserTerminalRouter.cpp
   src/terminal/TerminalClient.hpp
   src/terminal/TerminalClient.cpp
+  src/terminal/ServerFifoPath.hpp
+  src/terminal/ServerFifoPath.cpp
   src/terminal/SshSetupHandler.hpp
   src/terminal/SshSetupHandler.cpp
   src/terminal/UserTerminalHandler.hpp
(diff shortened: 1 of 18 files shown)


MisterTea:
LGTM, thanks!

Issue #515, opened 2022-04-30
`et` and `etserver` crash in Sentry on second launch (for Ubuntu 20.04)


jwmcglynn:
On second launch, et and etserver crash on my server, Ubuntu 20.04.  See attached lldb output:

```
(lldb) bt all
  thread #1, name = 'et'
    frame #0: 0x00007ffff7bfbdff libc.so.6`wait4 + 95
    frame #1: 0x0000555555a4e48f et`et::SubprocessToStringInteractive(command="ssh", args=size=2) at SubprocessToString.cpp:200:11
    frame #2: 0x000055555599a8d9 et`et::SshSetupHandler::SetupSsh([snip ...]) at SshSetupHandler.cpp:70:65
    frame #3: 0x000055555590ef09 et`main(argc=1, argv=0x00007fffffffe288) at TerminalClientMain.cpp:267:20
    frame #4: 0x00007ffff7b3d0b3 libc.so.6`__libc_start_main + 243
    frame #5: 0x000055555590504e et`_start + 46
  thread #2, name = 'sentry-http', stop reason = signal SIGSEGV: invalid address (fault address: 0x1000001b7)
    frame #0: 0x00007ffff79ce6f7 libssl.so.1.1`SSL_get_peer_certificate + 23
    frame #1: 0x00007ffff7f72179 libcurl.so.4`___lldb_unnamed_symbol814$$libcurl.so.4 + 265
    frame #2: 0x00007ffff7f762fe libcurl.so.4`___lldb_unnamed_symbol819$$libcurl.so.4 + 574
    frame #3: 0x00007ffff7f7714f libcurl.so.4`___lldb_unnamed_symbol853$$libcurl.so.4 + 95
    frame #4: 0x00007ffff7f22296 libcurl.so.4`___lldb_unnamed_symbol79$$libcurl.so.4 + 22
    frame #5: 0x00007ffff7f448cd libcurl.so.4`___lldb_unnamed_symbol329$$libcurl.so.4 + 1789
    frame #6: 0x00007ffff7f45a11 libcurl.so.4`curl_multi_perform + 145
    frame #7: 0x00007ffff7f3be8b libcurl.so.4`curl_easy_perform + 331
    frame #8: 0x0000555555b8e63f et`sentry__curl_send_task(_envelope=0x00005555561fe220, _state=0x00005555561f93e0) at sentry_transport_curl.c:192:19
[...]

PR #517, opened 2022-05-02
move sentry to vcpkg

MisterTea:


jwmcglynn:
I tested this locally on my Ubuntu 20.04 installation and am not seeing any crashes, I think this resolves #515

MisterTea:
Closes #515

PR #518, opened 2022-05-05
Always return a value from waitOnSocketData()

jubalh:
My guess is that the else part is what we will want to return if no
other case catches.

In any case we need a return value to prevent:
`error: control reaches end of non-void function [-Werror=return-type]`

### DIFF
--- a/src/base/Headers.hpp
+++ b/src/base/Headers.hpp
@@ -353,9 +353,8 @@ inline bool waitOnSocketData(int fd) {
     } else {
       FATAL_FAIL(selectResult);
     }
-  } else {
-    return FD_ISSET(fd, &fdset);
   }
+  return FD_ISSET(fd, &fdset);
 }
 
 inline string genRandomAlphaNum(int len) {


src/base/Headers.hpp
```diff
@@ -353,9 +353,8 @@ inline bool waitOnSocketData(int fd) {
     } else {
       FATAL_FAIL(selectResult);
     }
-  } else {
-    return FD_ISSET(fd, &fdset);
   }
+  return FD_ISSET(fd, &fdset);
```
jwmcglynn, line 357: So, looking at the original, it looks like the `FD_ISSET(fd, &fdset);` is still called, so both codepaths are correct.

I'm guessing that the compiler is confused because it doesn't know that the `FATAL_FAIL` on line 254 terminates the program.

This looks good to me, but a second opinion on my comment above would be great.

jubalh: Yes it's still called.
The difference is that now the compiler can easily see that there will always be an return value.

PR #522, opened 2022-06-11
fix readme regarding port forwarding

cbyrohl:
Adjusting the docs regarding port forwarding as

`user@machine:~$ et -t "2222:2222" remote`

works, while original readme's instructions fail with
```

user@machine:~$ et -t="2222:2222" remote
Exception: Argument ‘-t=2222:2222’ starts with a - but has incorrect syntax
```

Tested with et version 6.2.1


MisterTea:
Thanks!

PR #527, opened 2022-07-09
ssh-et: apply upstream updates

infokiller:
- Fixes SSH config not being applied to original remote
- Adds retries in case multiple processes are spawned
  concurrently (can happen with tmux-xpanes for example)


MisterTea:
Thanks!

PR #530, opened 2022-08-04
docs: write gpg key to `trusted.gpg.d` for APT

Rongronggg9:
`apt-key` has been deprecated since Debian 11 and will be removed in 12. Writing the GPG key to `/etc/apt/trusted.gpg.d` gets rid of this problem and works entirely the same as `apt-key add`.


MisterTea:
Makes sense, thanks!

PR #536, opened 2022-10-07
Support for ipv6 addresses (with or without port specified)

jshort:
Currently et does not support passing an ipv6 address as the host positional argument is assumed to be an ipv4 address or hostname with a single colon with the port after it and the rest is ignored, eg:

```
$ et 2620:10d:c083:1418:14ac:6bbc:4737:56ee                                                                                                                                     15:19:52  10.07.22  100% █
Could not reach the ET server: 2620:10
```

This change will support ipv4 address or hostname with port specified or ipv6 address with or without port specified.  Currently, if a port is specified in the positional arg, it overwrites what was specified (explicitly or by default) with the -p,--port option. I think this is backwards and may offer a PR to invert this logic.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,8 @@ jobs:
       - checkout
       - run:
           name: Set up ssh & known_hosts
-          command: sudo /etc/init.d/ssh start; rm -f ~/.ssh/id_rsa*; ssh-keygen -q -N "" -f ~/.ssh/id_rsa; sudo rm -f ~/.ssh/authorized_keys; cp ~/.ssh/id_rsa.pub ~/.ssh/authorized_keys; rm -f ~/.ssh/known_hosts; ssh -o "StrictHostKeyChecking no" localhost ls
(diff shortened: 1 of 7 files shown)


MisterTea:
This PR is in great condition. Can you modify the connect_to_initial_version circleci test to use ipv6 when connecting to the initial version?  Don't modify the address going the other way because version 6.0.0 won't have your commit.

jshort:
Sure! Note that this does not support ipv6 zero reduction (i.e. 0:0:0:0:0:0:0:1 -> ::1.  I'll file a task to implement that.

.circleci/config.yml
```diff
@@ -31,6 +31,9 @@ jobs:
       - run:
           name: Connect new -> old
           command: sudo ../root_version/build/etserver --daemon; sudo cp ../root_version/build/etterminal /usr/bin/etterminal; sleep 1; build/et -c "ls" localhost --logtostdout --verbose=9
+      - run:
+          name: Connect new -> old (ipv6)
+          command: sudo ../root_version/build/etserver --daemon; sudo cp ../root_version/build/etterminal /usr/bin/etterminal; sleep 1; build/et -c "ls" 0.0.0.0.0.0.0.1 --logtostdout --verbose=9
```
MisterTea, line 36: Shouldn't this be 0:0:0:0:0:0:0:1 (with colons instead of dots)?

jshort: Yep, just pushed fix! I also removed the starting of etserver and cp of etterminal since that is already done.

jshort: Looks like maybe the ssh daemon is not configured to work with ipv6: 

```
ssh: connect to host ::1 port 22: Connection refused
```

jshort:
Pushed a temp debug config for circle ci to see if ListenAddress :: is in the sshd config(s).

jshort:
@MisterTea 

Ok I got it working.  Sorry for those 16 force pushes!  I was running every tool I could think of to see why sshd was rejecting the ipv6 connection: netstat (gone in this version of Ubuntu), ss, ufw, ip(6)tables, lsof, ip addr, etc.  Turns out that sshd was running by default so the `/etc/init.d/ssh start` was a no-op so I had to change to `restart` to pick up my updated sshd_config 😭

PR #539, opened 2022-10-13
ipv6 abbreviated address support

### LINKED ISSUE #537: [Feature Request]Support collapsed zeroes format for ipv6 host addresses (i.e. ::1 for 0:0:0:0:0:0:0:1)
Now that the et terminal client supports ipv6 host positional args (with or without port specified), we should move the host argument parsing logic out of main() into a helper that has more powerful ipv6 parsing logic (not just counting 7 colons or 8 colons if a port is specified.

ipv4 ip or hostname will have at most 1 colon (if a port is specified).

ipv6 ip will have at least 2 colons (for a collapsed form with no port specified) and up to 8 if the address is fully expressed and a port is specified.

Ex:
* 1::1 => 1:0:0:0:0:0:0:1 (default port)
* [1::1]:2022 => 1:0:0:0:0:0:0:1 (port 2022) (need brackets though this format is not supported by OpenSSH
* 1:1:1:1:1:1::1 =>  1:1:1:1:1:1:0:1 (default port)

jshort:
This will support ipv6 addresses that are abbreviated with `::` such as
`::1` (for `0:0:0:0:0:0:0:1`) with some caveats around specifying the port
in the host positional argument.

Help output:

```
$> ./et --help
Remote shell for the busy and impatient
Usage:
  et [OPTION...] [user@]host[:port]

  Note that 'host' can be a hostname or ipv4 address with or without a port
  or an ipv6 address. If the ipv6 address is abbreviated with :: then it must
  be specfied without a port (use -p,--port).

  -h, --help                 Print help
      --version              Print version
  -u, --username             Username
      --host arg             Remote host name
  ...
```
Tests:

```
$> ./et ::1 -p 8080 --macserver
$> ./et 0:0:0:0:0:0:0:1 -p 8080 --macserver
$> ./et 0:0:0:0:0:0:0:1:8080 --macserver
$> ./et ::1 -p 8080 --macserver # fails because defaults to port 2022
```
Fixes #537

### DIFF
--- a/src/terminal/TerminalClientMain.cpp
+++ b/src/terminal/TerminalClientMain.cpp
@@ -38,60 +38,65 @@ int main(int argc, char** argv) {
   // Parse command line arguments
   cxxopts::Options options("et", "Remote shell for the busy and impatient");
   try {
-    options.positional_help("[user@]hostname[:port]").show_positional_help();
     options.allow_unrecognised_options();
+    options.positional_help("");
+    options.custom_help("[OPTION...] [user@]host[:port]\n\n"
+        "  Note that 'host' can be a hostname or ipv4 address with or without a port\n"
+        "  or an ipv6 address. If the ipv6 address is abbreviated with :: then it must\n"
+        "  be specfied without a port (use -p,--port)."
+    );
 
-    options.add_options()             //
-        ("h,help", "Print help")      //
-        ("version", "Print version")  //
-        ("u,username", "Username")    //
+    options.add_options()
+        ("h,help", "Print help")
+        ("version", "Print version")
+        ("u,username", "Username")
         ("host", "Remote host name",
-         cxxopts::value<std::string>())  //
-        ("p,port", "Remote machine port",
-         cxxopts::value<int>()->default_value("2022"))  //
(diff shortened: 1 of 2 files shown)


src/terminal/TerminalClientMain.cpp
```diff
@@ -38,60 +38,65 @@ int main(int argc, char** argv) {
...
+        "  or an ipv6 address. If the ipv6 address is abbreviated with :: then it must\n"
+        "  be specfied without a port (use -p,--port)."
+    );
+
+    options.add_options()
+        ("h,help", "Print help")
+        ("version", "Print version")
+        ("u,username", "Username")
         ("host", "Remote host name",
-         cxxopts::value<std::string>())  //
```
jshort, line 49: I removed all these trailing `//` as I'm not sure why there are there.  It looks like they are supposed to be `\` like when you extend a shell command in a script but they are not necessary in this context.

MisterTea: They are to avoid clang-format from putting multiple options per line

MisterTea:
Thanks!

Issue #547, opened 2022-11-18
etserver doesn't even start as a daemon


CmdQ:
I really would like to switch from `mosh` to `et`, but I cannot get it set up as a service. Compiling by hand failed with dependency issues that I could not resolve. I resorted to installing Homebrew on the machine and installing EternalTerminal through that, which worked.

But I don't get it running as a service or even as a daemon. The long version—in the begubber—reads like this:

```sh
[cmdq@AmazonLinux2]~% uname -a
Linux AmazonLinux2.us-west-2.amazon.com 5.4.214-134.408.amzn2int.x86_64 #1 SMP Wed Nov 2 05:27:06 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

[cmdq@AmazonLinux2]~% gdb --args /home/linuxbrew/.linuxbrew/bin/etserver --daemon --cfgfile=/etc/et.cfg --verbose=9
GNU gdb (GDB) Red Hat Enterprise Linux 8.0.1-36.amzn2.0.1
Copyright (C) 2017 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
BFD: warning: /home/linuxbrew/.linuxbrew/Cellar/et/6.2.1_1/bin/etserver: unsupported GNU_PROPERTY_TYPE (5) type: 0xc0008002
BFD: warning: /home/linuxbrew/.linuxbrew/Cellar/et/6.2.1_1/bin/etserver: unsupported GNU_PROPERTY_TYPE (5) type: 0xc0010001
BFD: warning: /home/linuxbrew/.linuxbrew/Cellar/et/6.2.1_1/bin/etserver: unsupported GNU_PROPERTY_TYPE (5) type: 0xc0010002
Reading symbols from /home/linuxbrew/.linuxbrew/bin/etserver...(no debugging symbols found)...done.
(gdb) run
[...]

PR #549, opened 2022-11-29
Remove daemon flag in systemd config

CmdQ:
For systemd, the command must start normally, not as daemon for it to work.


jshort:
Fixes #547. 👍

jshort (approved):

Issue #504, opened 2022-03-20
telemetry = false in et.cfg does not disable telemetry


fignew:
Hello,

When running
`$ et -v 7 --logtostdout 127.0.0.1`
I get the following output:
```
[INFO 2022-03-20 15:14:56,847 client-main ParseConfigFile.hpp:1400] unsupported config line:   ForwardX11 yes, ignored
[INFO 2022-03-20 15:14:56,847 client-main ParseConfigFile.hpp:1400] unsupported config line:   VisualHostKey yes, ignored
[INFO 2022-03-20 15:14:56,847 client-main ParseConfigFile.hpp:1181] /etc/ssh/ssh_config.d/*.confnot found
[INFO 2022-03-20 15:14:56,847 client-main TerminalClientMain.cpp:189] Parsed ssh config file, connecting to 127.0.0.1
[V1 2022-03-20 15:14:56,847 client-main TerminalClientMain.cpp:14] Connecting
[V4 2022-03-20 15:14:56,847 client-main TcpSocketHandler.cpp:83] Before selecting sockFd
[V4 2022-03-20 15:14:56,847 client-main TcpSocketHandler.cpp:87] sockFd 6is selected6
[INFO 2022-03-20 15:14:56,847 client-main TcpSocketHandler.cpp:107] Error connecting with 127.0.0.1: 111 Connection refused
[INFO 2022-03-20 15:14:56,848 client-main TcpSocketHandler.cpp:141] No host found
[V1 2022-03-20 15:14:56,848 client-main TerminalClientMain.cpp:17] Could not connect to host
Could not reach the ET server: 127.0.0.1:2022

Shutting down sentry
```
Seeing "Shutting down sentry" indicates to me that telemetry is still enabled

My et.cfg file is the following:
```
;######################################################################;
;                                                                      ;
;              THIS FILE IS MANAGED BY SALT - DO NOT EDIT              ;
;                                                                      ;
; The contents of this file are managed by Salt. Any changes to this   ;
; file may be overwritten automatically and without warning.           ;
;######################################################################;

[Networking]
port = 2022
# bind_ip = 0.0.0.0

[Debug]
verbose = 0
silent = 0
logsize = 20971520
telemetry = false
```

I've also tried telemetry = 0 and placing the option before and in the [Networking] section.

MisterTea:
The problem is that the print you are referring to is outside of the #define that disables telemetry:

[...]

PR #553, opened 2022-12-13
Parse telemetry ini config as boolean and make telemetry opt-in.

jshort:
In issue #504, it was brought to attention that while the command line arg for enabling telemetry was boolean, the ini config in et.cfg needed to be an int to change the value.  SimpleIni has a GetBoolValue function that treats the strings "0", "off", "no", "false" as boolean false and "1", "on", "true" as boolean true.

Also, this will make telemetry opt-in as opposed to opt-out as has been requested by multiple users.  The supplied et.cfg file will still have telemetry on.

### DIFF
--- a/etc/et.cfg
+++ b/etc/et.cfg
@@ -9,4 +9,4 @@ port = 2022
 verbose = 0
 silent = 0
 logsize = 20971520
-telemetry = 1
+telemetry = true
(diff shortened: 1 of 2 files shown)


jshort:
I'd say this has baked long enough without disagreement, merging...

PR #561, opened 2023-01-20
Add packaging information

jubalh:
To see which distributions package et.

We could also use a smaller badge like:
https://repology.org/badge/tiny-repos/eternalterminal.svg

What do you prefer?


MisterTea:
Wow, this is cool.  Thanks!

PR #562, opened 2023-01-21
Add BUILD_TESTING to make tests optional by choice

kruceter:
The option is enabled by default, so it should not disrupt the default behavior.


MisterTea:
Thanks!

PR #568, opened 2023-03-06
Fix build on CentOS 7

bin456789:
I got some compile errors when buiding ET on CentOS 7

```
- CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
- A suitable version of git was not found (required v2.7.4) and unable to automatically download a portable one. Please install a newer version of git.
- Can't locate IPC/Cmd.pm
- Can't locate Data/Dumper.pm
- Could NOT find Unwind (missing: Unwind_INCLUDE_DIR Unwind_LIBRARY Unwind_PLATFORM_LIBRARY)
- Could NOT find UTempter (missing: UTEMPTER_INCLUDE_DIR UTEMPTER_LIBRARIES)
- cannot find -latomic
```

Seems docs are too old for building ET on CentOS 7.
So I fix it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,13 +256,14 @@ Install dependencies:
 ```
 sudo yum install epel-release
 sudo yum install cmake3 boost-devel libsodium-devel ncurses-devel protobuf-devel \
-     protobuf-compiler gflags-devel protobuf-lite-devel libcurl-devel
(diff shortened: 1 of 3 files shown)


MisterTea:
Thanks so much!

PR #600, opened 2023-09-01
Force reverse tunnel socket to be binded to localhost

### LINKED ISSUE #587: Reverse tunnel listen interface
According to [this](https://github.com/MisterTea/EternalTerminal/issues/59#issuecomment-358578441) comment `anything that goes to localhost:4321 on the server will be received by 2224 on the local machine` but it definitely binds the socket to 0.0.0.0 instead of localhost and thus every outside connection can reach my local lemonade clipboard service while i definitely want it to be accessible only by the server i am connecting to.
E.x.
```bash
# on the client
-> et -r 58909:58909 $ip
# on the server
-> lsof -i -P -n | grep 58909
etserver  195783 root   14u  IPv4 2227673      0t0  TCP *:58909 (LISTEN)
```
while using plain ssh:
```bash
# on the client
-> ssh -R 58909:127.0.0.1:58909
# on the server
-> lsof -i -P -n | grep 58909
sshd      198448 boommy    8u  IPv6 3361053      0t0  TCP [::1]:58909 (LISTEN)
sshd      198448 boommy    9u  IPv4 3361054      0t0  TCP 127.0.0.1:58909 (LISTEN)
```
Notice that sshd listens for `127.0.0.0` and et listens for `*`.
Is it a bug, feature or requires an option to control the behavior?

lostystyg:
Fixes #587

### DIFF
--- a/src/base/TunnelUtils.cpp
+++ b/src/base/TunnelUtils.cpp
@@ -37,6 +37,7 @@ vector<PortForwardSourceRequest> parseRangesToRequests(const string& input) {
           int portRangeLength = sourcePortEnd - sourcePortStart + 1;
           for (int i = 0; i < portRangeLength; ++i) {
             PortForwardSourceRequest pfsr;
+            pfsr.mutable_source()->set_name("localhost");
             pfsr.mutable_source()->set_port(sourcePortStart + i);
             pfsr.mutable_destination()->set_port(destinationPortStart + i);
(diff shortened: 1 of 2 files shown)


MisterTea:
@lostystyg the CI is failing.  Can you try rebasing on latest master?

MisterTea:
nevermind, it's a bad test.  Thanks for the PR

PR #605, opened 2023-11-03
Add codespell support (config, CI action to detect new typos) + fix some typos

yarikoptic:


MisterTea:
Hey, I'm not sure why we are failing the clang-format test now, but can you fix those before merging?

yarikoptic:
I will submit later pr without changes to check if that check is green on general - https://github.com/MisterTea/EternalTerminal/actions/workflows/clang-format-check.yml shows that last time it was ran was 2 months ago, so may be just some new version of tooling causes it now to fail, not that changes in this pr

KapJI:
Clang-format is right here, extra letter made `{`s in that block unalighned. I guess you need to add an extra space for those line.

yarikoptic:
thanks, pushed results of `clang-format -i src/terminal/TelemetryService.cpp` -- let's see how it feels now!

MisterTea:
Thanks!

PR #614, opened 2024-01-31
Don't strip username from `jumphost`, add it if it's missing

### LINKED ISSUE #532: Using et through a jumphost
I've been using et for a while over a VPN - works great!

However, we are moving to use a jumphost instead.  I can not seem to get et to work through the jump host.

I can jump through the jumpiest just fine with ssh.  I do it like this:
```
ssh -A -J rayj@backdoor.dev.tilia-inc.com rayj@int.bullseye31.dev.tilia-inc.com
```

With et though things just hang for me.  Here is what I'm running:
```
et --jumphost backdoor.dev.tilia-inc.com  -c "tmux -CC -S /tmp/shareds attach -t shared || tmux -CC -S /tmp/shareds new -s shared" rayj@int.bullseye31.dev.tilia-inc.com
```

I'm probably doing something wrong.  So any help would be appreciated.  Or at least some guidance on how to debug this better?

ddebin:
Don't strip username from `jumphost`, add it if it's missing.

Fix a bug when connecting with two different usernames for server and jumphost.
Fix command line parsing bug when including username in jumphost.
Fix a bug when `ProxyJump` is set in `~/.ssh/config` with a username but without port.

Fix #532

### DIFF
--- a/src/terminal/SshSetupHandler.cpp
+++ b/src/terminal/SshSetupHandler.cpp
@@ -60,7 +60,7 @@ string SshSetupHandler::SetupSsh(const string& user, const string& host,
   if (!jumphost.empty()) {
     ssh_args = {
         "-J",
-        SSH_USER_PREFIX + jumphost,
+        jumphost,
     };
   }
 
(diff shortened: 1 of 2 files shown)


MisterTea:
Thanks for the PR! Disregard the CI failure: it's a vcpkg download issue.

PR #617, opened 2024-02-03
try system test

MisterTea:


src/terminal/TerminalClientMain.cpp
```diff
@@ -294,6 +297,11 @@ int main(int argc, char** argv) {
     }
 
     int jport = result["jport"].as<int>();
```
jshort, line 299: You can remove this to as it's not used.

PR #621, opened 2024-02-22
Fix devcontainer and update vcpkg for tests

Migelo:
Devcontainer: The original `ubuntu-21.04` image was failing to build, so I switched it to `debian-11` which works out of the box.

Tests: They were failing because `libwinpthread < 10.0` is not available on the `https://repo.msys2.org` anymore. See for example the 404 not found errors [here](https://github.com/MisterTea/EternalTerminal/actions/runs/7741605598/job/21108940236#step:12:165).

So what I did for testing is I updated `vcpkg` one month at the time until the tests started passing and ended up on vcpkg commit [4a3c366](https://github.com/microsoft/vcpkg/commit/4a3c366f2d0d0eaf034bfa649124768df7cfe813).


Migelo:
I guess this supersedes https://github.com/MisterTea/EternalTerminal/pull/618.

MisterTea:
Nice!  I'll want to upgrade vcpkg further in #618 which will require making some changes in et, but this is a great PR.  Thanks a lot!

PR #622, opened 2024-02-25
update actions and docker images

Migelo:
The only non-trivial change is updating `actions/upload-artifact` from `v2` to `v4`, [source](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes). They don't allow for artifacts with the same name being uploaded and overwriting each other. This happened in the linux-ci workflow where different gcc versions were used but the artifact they produced had the same name, making the `v4` action fail. I fixed it by creating a unique name for each gcc version.

        -  name: et-client-${{matrix.os}}
        +  name: et-client-${{matrix.os}}-gcc${{matrix.gcc}}

PR #623, opened 2024-02-25
add renovatebot

Migelo:
As discussed in https://github.com/MisterTea/EternalTerminal/pull/622 this pull request adds the [renovatebot](https://docs.renovatebot.com/) to this repo. 

The current config will automatically merge minor updates and patches, major version bumps need to be merged manually.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,40 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:base"
+  ],
+  "includeForks": true,
+  "lockFileMaintenance": {
+    "enabled": true,
+    "automerge": true,
+    "automergeType": "pr",
+    "platformAutomerge": true
+  },
+  "platformAutomerge": true,
+  "packageRules": [
+    {
+      "description": "Automerge non-major updates",
+      "matchUpdateTypes": [
+        "minor",
+        "patch"
+      ],
+      "enabled": true,
+      "automerge": true,
+      "automergeType": "pr",
+      "platformAutomerge": true
+    },
+    {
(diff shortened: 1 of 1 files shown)


MisterTea:
Can you add a github action to run renovate? https://github.com/renovatebot/github-action

Migelo:
Something that runs every 15 minutes like this? Happy to do it!

```
name: Renovate
on:
  schedule:
    # The "*" (#42, asterisk) character has special semantics in YAML, so this
    # string has to be quoted.
    - cron: '0/15 * * * *'
jobs:
  renovate:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v4.1.1
      - name: Self-hosted Renovate
        uses: renovatebot/github-action@v40.1.1
        with:
          configurationFile: example/renovate-config.js
          token: ${{ secrets.RENOVATE_TOKEN }}
```

I never knew one can run renovate as an action, I've always just let it run on its own accord, what benefits do you recon this brings?

MisterTea:
"let it run on its own accord"

What does this mean?  I assumed that without a GitHub action I would need to run some command regularly to get updates.

Migelo:
It just triggers every so often on its own. Without a schedule I guess it runs when the load on the bot is low? After adding renovate.json you simply need to enable https://github.com/apps/renovate for this repo and that is it, no gh action needed.

MisterTea:
Oh!  That's the part I was missing. Ok, with the GitHub hook we don't need an action.

Migelo:
Great. Then the only thing you need to do before merging this is to add the renovate GitHub app I linked above to this repo.

MisterTea:
Thanks!!

Issue #548, opened 2022-11-29
(Relatively) high CPU use


xxxserxxx:
`etserver` is unexpectedly busy, even when no clients are connected. While the CPU use is not pegging the CPU, the fact that it is often at the top of the process list on a mostly quiescent VPS is curious. What is it doing?

Related: #11

The baseline comparison is `mosh`. Mosh is on-demand, so when there are no clients, `mosh` server is not running. When a client **is** connected, `mosh` server rarely (if ever) consumes enough CPU to enter the top 20. The only other non-standard services running on this server are `wireguard` and `sshguard`. There is no graphical environment, so other load is light. 

In contrast, etserver regularly sits at the top of the CPU usaeage list at around ca 0.7% - 1.0%, over `top`  and even the `sshguard`. This is with no clients. If a client connects, then `etterminal` is started and `etserver` jumps to a consistent 1.0% - 1.3% in addition to 1.0% for `etterminal`. With a connection eternalterminal processes consumes between 2-2.5% CPU use when idle.

Compared to running a rather busy terminal program, `gotop` with a 500ms refresh in `mosh` (same server, different connection), `etserver` remains the top process. Applications with high drawing demands -- `ncurses`, etc -- do not increase either the `etserver` or `etterminal` loads, so this CPU use appears to be baseline background busywork.

In the logs for `etserver` are only the starting and listening messages, until a client connects; after a client connection are a few connection set-up `INFO` messages, no errors, and then a series of keep alive messages. The client log (on the server) is even more quiet, with merely a connection message and an "is a good socket" confirmation. The server at `-v 9` reports no more information (with no connection) than at verbosity 0: literally, just 4 start-up `INFO` messages.

- Arch 5.15.64-1-lts
- etserver 6.2.1
- [etserver-2022-11-29_08-28.log](https://github.com/MisterTea/EternalTerminal/files/10114128/etserver-2022-11-29_08-28.log)
- [etterminal-ser-tiEplprDIqn7QMjr.log](https://github.com/MisterTea/EternalTerminal/files/10114129/etterminal-ser-tiEplprDIqn7QMjr.log)
- [et.cfg.txt](https://github.com/MisterTea/EternalTerminal/files/10114138/et.cfg.txt) (renamed to .txt for github)

1. Why is etserver so busy (~1%) when it is idle, and with no client connections?
2. Why are the pair (etserver and etterminal) so busy (~2-2.5% combined) *with* a client connection?

[...]

PR #624, opened 2024-02-27
increase timeout to address (Relatively) high CPU use #548

Migelo:
Initial implementation to try and address the high cpu usage discussed in #548.


MisterTea:
Thanks!!

MisterTea:
Closes #548

Issue #357, opened 2020-10-31
et error after connecting to server


RonanMacF:
Port 2022 is not usable on the server so I am trying to connect using different ports. I have confirmed that et is installed on both the client and server and is visible in PATH. I have paswordless login setup on the server also. I have identified open listening ports using `netstat -tnlp` and can see the following
```
tcp        0      0 0.0.0.0:5914            0.0.0.0:*               LISTEN      -
tcp6       0      0 :::6014                 :::*                    LISTEN      -
```

When I try to connect to the server using either one of these ports I get the following.
```
[INFO 2020-10-31 16:49:14,306 client-main ParseConfigFile.hpp:1378] /Users/ronan/.ssh/confignot found
[INFO 2020-10-31 16:49:14,307 client-main ParseConfigFile.hpp:1353] unsupported config line: 	SendEnv LANG LC_*
, ignored
[INFO 2020-10-31 16:49:14,307 client-main TerminalClientMain.cpp:172] Parsed ssh config file, connecting to hostname
[INFO 2020-10-31 16:49:14,494 client-main TcpSocketHandler.cpp:88] Connected to server: hostname using fd 6
zsh:2: command not found: etterminal
Error starting ET process through ssh, please make sure your ssh works first
```

I confirmed ssh is working on the server and it looks like the connection suceeded.

MisterTea:
"zsh:2: command not found: etterminal"

That's the issue.  the etterminal command is not in /usr/bin .  Updates to the path in bashrc will not work because the etterminal command runs before that.

RonanMacF:
> "zsh:2: command not found: etterminal"
> 
> That's the issue. the etterminal command is not in /usr/bin . Updates to the path in zshrc will not work because the etterminal command runs before that.

appreciate it.
```
$ whereis et
et: /usr/bin/et /usr/include/et /usr/share/et /nix/store/yyaq9iy2v6sbsm5631s1n70zp0a9m5l1-user-environment/bin/et
```

then trying to connect again using `et ronan@host -p xx -v 9 --logtostdout` where xxx is the mix of the open ports found in the original message.
```
[INFO 2020-11-05 19:46:05,795 client-main TerminalClientMain.cpp:172] Parsed ssh config file, connecting to host
[V1 2020-11-05 19:46:05,795 client-main TerminalClientMain.cpp:11] Connecting
[V4 2020-11-05 19:46:05,980 client-main TcpSocketHandler.cpp:75] Before selecting sockFd
[...]

PR #629, opened 2024-04-10
fix: etterminal_path is the incorrect option name

fennecfx:
Caused by https://github.com/MisterTea/EternalTerminal/commit/c1d54e71304fe435eb7bf14bcfa62cf783100314

Should fix https://github.com/MisterTea/EternalTerminal/issues/357#issuecomment-1452357927


MisterTea:
Thanks!

PR #642, opened 2024-06-03
Increase socket timeout to 30s

MisterTea:


jshort:
I wonder if it is a the timeout in waitOnSocketData:

```
inline bool waitOnSocketData(int fd) {
  fd_set fdset;
  FD_ZERO(&fdset);
  FD_SET(fd, &fdset);
  timeval tv;
  tv.tv_sec = 1;
  tv.tv_usec = 0;
  VLOG(4) << "Before selecting sockFd";
  const int selectResult = select(fd + 1, &fdset, NULL, NULL, &tv);
```

Which is 1 sec, and select is repeatedly returning 0 which indicates a timeout.

PR #651, opened 2024-08-14
[website] Fix the broken GitHub ribbon image

bryant1410:
The image link is broken. I changed it to the same one (which seems to be the most official one), and it is the same size. See https://web.archive.org/web/20210107045132/https://eternalterminal.dev/ for a reference.


MisterTea:
Thanks so much!  That was super annoying before you fixed it.

PR #658, opened 2024-10-02
Update Dockerfiles to use a modern distro and the offical package

coandco:
Currently, the Dockerfiles in the offical repo are non-functional, because CentOS is no longer a thing and the official package repository has been taken offline.  As such, this pull request updates them to use Debian 12 and uses the official EternalTerminal repository to install the ET binaries into the container.


MisterTea:
Thanks!

PR #672, opened 2024-11-16
Speedup build process

### LINKED ISSUE #586: Dependency/Readme clean up
Assuming my quick check of the build dependencies is correct, could the main page be updated to reflect the current dependencies? It would help with cleaning up distro packages and could save build time for source build systems.
## Deprecated dependencies
I believe the following libraries are no longer required for the latest builds:
`ncurses` (since v3 on, I believe the last usage was [v2.1.0 NCursesOverlay.hpp](https://github.com/MisterTea/EternalTerminal/blob/6d2984cacd2d8999acf85dd3f38212341c3c4cdc/terminal/NCursesOverlay.hpp#L6), [v2.1.0 CMakeLists.txt](https://github.com/MisterTea/EternalTerminal/blob/et-v2.1.0/CMakeLists.txt))
`gflags` (since v6 on, last usage was on v5, with removal here: #206)

## Potentially optional dependencies:
`boost` When looking at the curent master https://github.com/MisterTea/EternalTerminal/commit/2e06248f75d40042d3b116b233bce58db5022e6d, I'm not seeing much usage of boost (most (or all) are checks to see if it is availible); when grep'ing through the v6 releases, I'm seeing a lot of references to it in the [`external`](https://github.com/MisterTea/EternalTerminal/tree/2e06248f75d40042d3b116b233bce58db5022e6d/external) and the more recent (since v6.0.4, but not in master; presumably for vcpkg) `external_imported` (different commit 
[5eb70c/external_imported](https://github.com/MisterTea/EternalTerminal/tree/5eb70c2169870394aad2fe67a42a2eabf519232f/external_imported)) folders. Additionally, I am able to build a few random v6 releases without boost; what is lost by not including this library in a build and could the main page be updated to make this an optional include (with some details on why one may/may not want to include it (is this library currently included solely for vcpkg?))?

timsu92:
This PR speeds up building process by removing unnecessary packages, shrink download size and compile parallely.

I've checked that packages listed in #586 is correct. Removing `libboost-dev` and `libgflags-dev` still build successfully in Ubuntu 24.04.

close #586

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,8 +227,8 @@ Grab the deps and then follow this process.
 Debian/Ubuntu Dependencies:
 
 ```
-sudo apt install libboost-dev libsodium-dev autoconf libtool \
-	libprotobuf-dev protobuf-compiler libgflags-dev libutempter-dev libcurl4-openssl-dev \
+sudo apt install libsodium-dev autoconf libtool \
+	libprotobuf-dev protobuf-compiler libutempter-dev libcurl4-openssl-dev \
     build-essential ninja-build cmake git zip
 ```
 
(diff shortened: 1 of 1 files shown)


MisterTea (approved):
Thanks!!

PR #674, opened 2024-11-18
Include systemd files in the final built package

timsu92:
This allows built package to install systemd service file and `/etc` config file.

Files in `debian` folder is used to build up and tear down systemd service. Those files were generated by `debhelper`, `dh-cmake` and related packages.


timsu92:
@jgcircuit I'm ready~
I just found that there was a git conflict~

MisterTea:
Thanks!

PR #675, opened 2024-11-19
updated install instructions for debian

### LINKED ISSUE #607: APT-KEY is deprecated, update instructions for install
As you may or may not know, apt-key is deprecated. I did the below instead and it works fine for me. I think Ubunutu would be the same or similar, but I don't know.

I used bits from [this](https://medium.com/@glegoux/ubuntu-22-04-jammy-jellyfish-apt-key-is-deprecated-2dbbee8aec84) article to figure it out. I didn't bother migrating.

### OS Info
cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian

Distro: DietPi v8.23.3
Device model : RPi 4 Model B (aarch64)

### Step 1: Fix /etc/apt/sources.list.d/et.list

Add arch signed by with the gpg key location.

The gpg file should go in /etc/apt/keyrings/, /usr/share/keyrings, or really anywhere as long as it matches what is in the et.list.

The format should be something like this:
```
deb [arch=arm64 signed-by=/etc/apt/keyrings/et.gpg] https://mistertea.github.io/debian-et/debian-source/ bookworm main
```

You could probably automate the arch field using 

> dpkg --print-architecture

and at least on my version of debian, to get the codename:

> cat /etc/os-release | sed -n -e 's/^VERSION_CODENAME=//p'

I figure it isn't great or worth it to automate the codename part, but you could insert the architecture like this:

> echo "deb [arch="$(dpkg --print-architecture) "signed-by=/etc/apt/keyrings/et.gpg] https://mistertea.github.io/debian-et/debian-source/ bookworm main" | sudo tee /etc/apt/sources.list.d/et.list

### Step 2: Put the GPG file in the "correct" location

The gpg file should go in /etc/apt/keyrings/, /usr/share/keyrings, or really anywhere as long as it matches what is in the et.list

```
curl -sS https://mistertea.github.io/debian-et/et.gpg -o /etc/apt/keyrings/et.gpg
```

After that, I just ran 'apt update' and 'apt install et' like normal.

stonerl:
closes #607

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,6 @@ sudo cp ../init/launchd/homebrew.mxcl.et.plist /Library/LaunchDaemons/homebrew.m
 sudo launchctl load -w /Library/LaunchDaemons/homebrew.mxcl.et.plist
 ```
 
-
 Alternatively, a package is available in MacPorts:
 
 ```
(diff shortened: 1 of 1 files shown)


MisterTea (approved):
Thanks!

stonerl:
You're welcome. Is the website's code also hosted somewhere on GitHub? I would open a PR to correct the installation instructions there as well.

MisterTea:
@stonerl yes it's in the gh-pages branch

PR #687, opened 2025-03-12
Fix multiple options accepted for single-value options

p12tic:
Currently single-value options such as --tunnel will accept multiple values, e.g. --tunnel "<...>" --tunnel "<...>". Confusingly for the user only the last value will be used without any warning.

This commit fixes this by emitting an error message and exiting the program if multiple option values were supplied when single value was expected.

There are several more cases where this same approach could be applied, but these are not as trivial to map to the newly added function (e.g. `result["terminal-path"].as<string>()`), so I left them as is.

### DIFF
--- a/src/terminal/TerminalClientMain.cpp
+++ b/src/terminal/TerminalClientMain.cpp
@@ -29,6 +29,23 @@ void handleParseException(std::exception& e, cxxopts::Options& options) {
   exit(1);
 }
 
+template <class T, class DefaultT>
+T extractSingleOptionWithDefault(const cxxopts::ParseResult& result,
+                                 const cxxopts::Options& options,
+                                 const string& name, DefaultT defaultValue) {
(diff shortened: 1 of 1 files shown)


p12tic:
CirctleCI tests fail because "Job was rejected because resource class medium, image ubuntu-2004:202101-01 is not a valid resource class ".

MisterTea:
Thanks!

MisterTea:
@p12tic your PR is failing CI. Can you take a look?  The circle CI tests are broken but the others should pass.

p12tic:
@MisterTea Should be good now. Sorry for wasting your time with the first iteration, I should have been more careful.

p12tic:
@MisterTea Just a friendly ping :) Thanks a lot for your time.

MisterTea:
Thanks!!

PR #689, opened 2025-04-18
Added support for glob patterns (* and ?) in SSH config Include directive parser.

ccanoe:
I added support for glob patterns (* and ?) in the Include directive of the SSH config parser used by Eternal Terminal.

This enables the inclusion of multiple SSH config files using wildcard patterns, matching OpenSSH behavior for directives like:

Include ~/.ssh/config.d/*.conf
Include host_?.conf

Tested locally with various pattern combinations and confirmed correct parsing and connection behavior.


MisterTea:
@ccanoe if you rebase, this should be good now

MisterTea:
Thanks!

PR #700, opened 2025-08-04
Avoid microsecond format string parsing

jamestaylr:
Avoid microsecond format string parsing

Fixes a bug in the implementation of https://github.com/MisterTea/EternalTerminal/pull/680. Log creation uses [`std::strftime`](https://en.cppreference.com/w/cpp/chrono/c/strftime) which does not support the `%f` format string. Instead, build the microseconds outside of `strftime` and append it after.

After the fix, the file names are now correclty appending microseconds (after testing):

```
test_prefix-2025-08-03_21-37-11.810495
test_prefix-stderr-2025-08-03_21-37-11.810495
```

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/MisterTea/EternalTerminal/pull/700).
* __->__ #700
* #680

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,7 +297,7 @@ Alternatively, open the file /etc/systemd/system/et.service in an editor and cor
 
 
 ```
-ExecStart=/usr/local/bin/etserver --cfgfile=/etc/et.cfg
+ExecStart=/usr/local/bin/etserver --cfgfile=/etc/et.cfg --logtostdout
 ```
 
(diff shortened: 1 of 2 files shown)


jamestaylr:
Before merging this, I want to do some more local testing. Will comment here after I get the chance to do that.

PR #705, opened 2025-10-20
Fix ProxyJump to support SSH ports and host aliases

andrewleech:
ET currently fails to connect through jump hosts when the jump host uses a non-standard SSH port or when ProxyJump references an SSH config Host alias.

For example, with this SSH config:
```ssh
Host jumphost
  HostName jump.example.com
  Port 2222

Host destination
  HostName dest.example.com
  ProxyJump jumphost
```

Running `et destination` fails with `Could not reach the ET server: jumphost:2022`

The issue is that the ProxyJump parsing code was stripping SSH ports from the jump host value:
```cpp
if (colonIndex != string::npos) {
  jumphost = proxyjump.substr(0, colonIndex);  // Strips :2222
}
```

Additionally, Host aliases like "jumphost" weren't being resolved to actual hostnames, so ET tried to connect to the alias name instead of the configured hostname.

This PR fixes both issues by:
- Preserving the full ProxyJump value `[user@]host[:port]` when passing to SSH via `-J` flag
- Extracting just the hostname from the ProxyJump value for ET socket connections
- Looking up Host aliases in SSH config to get the actual hostname

Tested with the config above. Before the fix `et destination` fails, after it successfully connects via jump.example.com:2222 (SSH) -> destination (ET).

No breaking changes expected - existing ProxyJump configurations should continue to work, though I don't currently have a working jumphost setup to verify backwards compatibility directly.

### DIFF
--- a/src/terminal/HostParsing.hpp
+++ b/src/terminal/HostParsing.hpp
@@ -1,0 +1,58 @@
+#ifndef __ET_HOST_PARSING__
+#define __ET_HOST_PARSING__
+
+#include "Headers.hpp"
+
+namespace et {
+
+// Parsed components of a [user@]host[:port] string
+struct ParsedHostString {
+  string user;
+  string host;
+  string portSuffix;  // includes colon, e.g. ":22"
+};
+
+// Parse a host string in [user@]host[:port] format
+// Handles IPv6 addresses in bracket notation: [::1], [::1]:22, user@[::1]:22
+inline ParsedHostString parseHostString(const string& hostString) {
+  ParsedHostString result;
+  string remaining = hostString;
+
+  // Extract user@ prefix if present
+  size_t atIndex = remaining.find("@");
+  if (atIndex != string::npos) {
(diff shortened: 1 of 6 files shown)


andrewleech:
Found the hang. Test 4 in connect_with_jumphost.sh was missing --port 9901 for the destination ET server.

Without it, `destinationPort` defaults to 2022. The jump daemon starts on the jumphost with `--dstport=2022` but the destination etserver is at 9901, so `UserJumphostHandler::connect()` fails. The jump daemon then enters its reconnect loop indefinitely. Since it's daemonized (via `DaemonCreator::createSessionLeader()`) it doesn't block the SSH subprocess, so `SubprocessToStringInteractive` returns fine with the IDPASSKEY, but the et client then hangs in `TerminalClient::run()` waiting for data through the jump channel that never arrives.

I've pushed a fix - just adds `--port 9901` to the Test 4 et command. ctest passes locally (259s). Should unblock CI.

MisterTea:
Thanks!

PR #706, opened 2025-10-20
fix circleci

MisterTea:


jshort:
Might need some updated sshd config for the new version, will debug later.

jshort:
Merged in #714

Issue #711, opened 2025-11-21
Crash due to null pointer in jumphost forwarding loop (race condition)


tooji:
## Summary
On multi-tenant bastion hosts, etterminal can crash with SIGABRT when forwarding client packets to a destination login node before the bastion→login connection is fully established. This is caused by a race condition: packets from the client arrive and are forwarded via `UserJumphostHandler` before `ClientConnection`'s `writer` member is initialized, leading to a null pointer dereference (`BackedWriter::write(this=0x0)`).

## Stack Trace
```
#13 et::BackedWriter::write (this=0x0, packet=...) at src/base/BackedWriter.cpp:15
#14 et::Connection::write (this=0xaaaad592a930, packet=...) at src/base/Connection.cpp:180
#15 et::Connection::writePacket (this=0xaaaad592a930, packet=...) at src/base/Connection.cpp:53
#16 et::UserJumphostHandler::run (this=0xffffe43344b8) at src/terminal/UserJumphostHandler.cpp:168
```

## Reproduction
1. Connect through a busy bastion: `et --jumphost user@bastion --dsthost=login.server.com --dstport=2022`
2. Begin sending data (type or paste in terminal) immediately on connect
3. If client data arrives before bastion→login connection is established, a crash occurs

## Code Pointer
- Forwarding loop in `src/terminal/UserJumphostHandler.cpp` around line 168:
```cpp
if (routerSocketHandler->readPacket(routerFd, &p)) {
    jumpclient->writePacket(p); // CRASH: writer not initialized yet
}
```

## Proposed Fix
Add a null check for `jumpclient->getWriter()` before invoking `writePacket()`. Alternatively, block forwarding until the bastion→login connection is ready.

Label: bug

PR #715, opened 2025-12-03
Remove null pointer deref in connection read and write primitives (#711)

alichtman:
See bug report and full analysis at #711.

Add null pointer guards in `Connection::read()` and `Connection::write()` to prevent crashes when reader/writer are accessed before initialization.


alichtman:
@jshort: This is a variation on the fix proposed in https://github.com/MisterTea/EternalTerminal/issues/711#issue-3652939005

Instead of adding the null check before invoking `writePacket()`, I added it one layer down, in `write`. I also applied the same fix to `read`.

PR #717, opened 2025-12-05
Update utempter designation to be 'etterminal' plus misspelling update.

jshort:
The actual process that defines the wtmp session is etterminal, so let's make that very clear in the utempter record.

ex before:

```
❯ last | head
jwshort  pts/8        et [275430]      Fri Dec  5 10:20    gone - no logout
```

ex after:

```
❯ last -w | head
jwshort  pts/8        etterminal [4036829]     Fri Dec  5 10:03 - 10:20  (00:17)
```

### DIFF
--- a/src/terminal/PseudoUserTerminal.hpp
+++ b/src/terminal/PseudoUserTerminal.hpp
@@ -22,9 +22,9 @@
 #include "UserTerminal.hpp"
 
 namespace et {
-class PsuedoUserTerminal : public UserTerminal {
+class PseudoUserTerminal : public UserTerminal {
  public:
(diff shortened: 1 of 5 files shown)


src/terminal/PseudoUserTerminal.hpp
```diff
@@ -46,7 +46,7 @@ class PsuedoUserTerminal : public UserTerminal {
 #ifdef WITH_UTEMPTER
     {
       char buf[1024];
-      sprintf(buf, "et [%lld]", (long long)getpid());
```
jshort: This is the only material change in this PR.

PR #726, opened 2025-12-15
Fix errno corruption causing spurious ENOENT crashes in etterminal

ejc3:
## Summary

Fixes a bug where `etterminal` crashes with misleading `ENOENT` ("No such file or directory") errors when the terminal session ends or encounters read errors. The crash prevents session resumption, defeating ET's core purpose.

## The Bug

In `UserTerminalHandler::runUserTerminal()`, errno was being read **after** `VLOG()` calls:

```cpp
int rc = read(masterFd, b, BUF_SIZE);
VLOG(4) << "Read from terminal";  // ← This can modify errno!
FATAL_FAIL(rc);                   // ← Reports corrupted errno
```

When `read()` failed (e.g., with `EIO` when the shell process died), the `VLOG()` call would perform logging I/O operations that could set errno to `ENOENT` (e.g., from a log directory lookup). `FATAL_FAIL` would then crash with the wrong error message.

**Observed symptoms:**
- `etterminal` crashes with `FATAL ... Error: (2): No such file or directory`
- Sessions cannot be resumed after network disconnections
- Crash appears to be PTY-related but `ENOENT` is impossible from `read()` on a PTY

**Root cause:** The `ENOENT` was coming from the logging subsystem, not from the PTY read.

## The Fix

1. **Save errno immediately** after `read()` before any logging operations
2. **Handle errors gracefully** instead of using `FATAL_FAIL`:
   - `rc > 0`: Success, process data
   - `rc == 0`: EOF, clean session end
   - `rc == -1` with `EAGAIN`/`EWOULDBLOCK`: Transient, retry
   - `rc == -1` with other errors: Log correct error, clean shutdown

This matches the error handling pattern already used in `TerminalServer.cpp` (lines 296-317).

## Changes

- `masterFd` read: Save errno immediately, move VLOG after success check, handle all error cases gracefully with proper cleanup
- `routerFd` read: Save errno immediately, handle `EAGAIN`/`EINTR` as transient, throw with correct errno for fatal errors

## Testing

- Built and tested on Ubuntu 24.04 ARM64
- Verified sessions end gracefully instead of crashing
- Error messages now show correct errno values

### DIFF
--- a/src/terminal/UserTerminalHandler.cpp
+++ b/src/terminal/UserTerminalHandler.cpp
@@ -140,7 +140,8 @@ void UserTerminalHandler::runUserTerminal(int masterFd) {
           if (readErrno == EAGAIN || readErrno == EINTR) {
             continue;  // Transient error, retry
           }
-          throw std::runtime_error(string("Router read error: ") + strerror(readErrno));
+          throw std::runtime_error(string("Router read error: ") +
+                                   strerror(readErrno));
         }
         if (rc == 0) {
           throw std::runtime_error(


ejc3:
## Testing Results

Deployed the fix to a production server and confirmed it resolves the issue.

### Before the fix (crashes with misleading ENOENT):
```
[FATAL 2025-12-14 17:28:17 UserTerminalHandler.cpp:107] Stack Trace: 
Error: (2): No such file or directory

[FATAL 2025-12-14 20:49:01 UserTerminalHandler.cpp:107] Stack Trace: 
Error: (2): No such file or directory

[FATAL 2025-12-15 08:10:31 UserTerminalHandler.cpp:107] Stack Trace: 
Error: (2): No such file or directory
```

### After the fix (sessions survive disconnects):

Connection resets occurred but sessions stayed alive:
```
16:39:14 - Connection reset by peer → No crash ✓
16:43:44 - Connection reset by peer → No crash ✓
16:43:46 - Connection reset by peer → No crash ✓
17:34:38 - Connection reset by peer → No crash ✓
```

Session that experienced 3 connection resets is still running:
```
$ ps aux | grep etterminal
ubuntu   1622845  0.0  0.0  17524  4204 ?  Ss   16:06   0:06 etterminal --verbose=1
```

The errno corruption was confirmed - the `ENOENT` errors were coming from the logging subsystem (VLOG), not from the PTY read. After saving errno immediately, the sessions handle disconnects gracefully instead of crashing.

jshort:
This is pretty similar to @vif's PR that fixed MacOS etserver disconnects in #719

jshort:
I believe so, it's just more about rc and errno handling when we are `read`ing and there are multiple places in our binaries where this occurs.

PR #730, opened 2026-01-11
Add opt-in flow control: --flow-control {backpressure,discard} (#631)

### LINKED ISSUE #631: Flow Control
I've noticed that ET struggles with Flow Control. If you run

`cat /dev/zero | base64 -w 0`

it is not possible to send a ^C to the server. Using SSH directly it works fine.

bmaurer:
## Problem (#631)

When a process produces output faster than the network can deliver it, the excess queues in kernel buffers — mostly the server's TCP send buffer, which autotunes to many MB. On a saturated link the display falls steadily behind real time and Ctrl-C appears dead: the interrupt is delivered, but the prompt has to wait behind the entire queue. On a 100KB/s test link, Ctrl-C took **56–116 seconds** to take effect.

## Design: strictly opt-in flow control

This revision makes flow control a per-session mode chosen by the client, with the default being **no change at all**:

- **`--flow-control none` (default)** — the existing code path, unchanged. The proto field isn't even serialized when unset, so new/old clients and servers interoperate exactly as before. Users who don't opt in are unaffected.
- **`--flow-control backpressure`** — terminal output is staged in a small (64KB) buffer drained only when `select()` reports the socket writable; when it fills, the server stops reading the PTY and the remote process pauses, like plain ssh. **Lossless** — safe for consumers that need every byte, e.g. `tmux -CC`.
- **`--flow-control discard`** — same buffer, but when full the *oldest* chunk is dropped. The remote process **never stalls** — even while the client is disconnected — and the display stays near real time. Output that exceeded the link rate is missing from scrollback.

### Kernel buffer tuning (opt-in sessions only)

Application-level buffering only helps if the backlog actually waits in the application buffer, so opted-in sessions also shrink the kernel queues (`SocketHandler::minimizeKernelBuffering`, a no-op by default):

- `TCP_NOTSENT_LOWAT=32KB` on the client connection: `select()` reports writable only when <32KB is unsent, keeping the backlog in the WriteBuffer where it can be gated or dropped. In-flight (sent-but-unacked) data is not limited, so throughput on high-BDP links is unaffected. Reapplied each loop iteration because reconnects replace the socket (and fd numbers get reused).
- `SO_SNDBUF=64KB` on the etterminal→etserver unix socket (a pure queue with no in-flight component, so clamping costs nothing locally). etterminal learns the mode via a new `TermInit` field.

In the jumphost path, discard only drops `TERMINAL_BUFFER` packets; control packets (port forwarding, responses, keepalives) are always delivered, and reads pause when the queue is dominated by them so it stays bounded.

## Measured results (100KB/s throttled link, 2.7MB/s producer)

| mode | display lag @30s | producer | Ctrl-C → prompt |
|---|---|---|---|
| `none` (unchanged) | ~31s, growing ~1s/s | throttled | 56–116s |
| `backpressure` | ~2.5s, bounded | throttled (lossless) | 2–3s |
| `discard` | ~1.2s, bounded | **full speed** | **2.1s** |

With `discard`, the producer also runs at full speed through a 60s network outage, and on reconnect the display recovers to ~1.2s of lag instead of replaying the backlog.

## Commits

1. **Fix busy-loop when a socket closes during waitOnSocketData** — standalone bugfix: EBADF/EINVAL from `select()` was treated like EINTR, so `readAll` retry loops spun at 100% CPU on a concurrently-closed fd.
2. **Add opt-in flow control** — the feature itself, with unit tests (`WriteBufferTest`) and an integration test covering both opt-in modes end-to-end (`ServerFlowControlDataTransferTest`).
3. **Add E2E flow-control test harness** — throttle-proxy harness measuring display lag, producer stalls, and Ctrl-C latency per mode, with disconnect simulation; methodology and full results in `test/e2e/REPORT.md`.

Fixes #631

### DIFF
--- a/proto/ETerminal.proto
+++ b/proto/ETerminal.proto
@@ -57,10 +57,25 @@ message PortForwardData {
   optional bool closed = 5;
 }
 
+enum FlowControlMode {
+  // Legacy behavior: no application-level buffering or kernel buffer
+  // tuning. This is the default and what servers assume for old clients
+  // that don't send flow_control_mode: current users are unchanged.
+  FLOW_CONTROL_NONE = 0;
+  // Bounded buffering; when full, stop reading from the source. Lossless
+  // (like plain ssh, but with small, tuned buffers so Ctrl-C stays
+  // responsive on a saturated link).
+  FLOW_CONTROL_BACKPRESSURE = 1;
+  // Bounded buffering; when full, drop the oldest pending output. The
+  // producing process never stalls and the display stays near real time.
+  FLOW_CONTROL_DISCARD = 2;
+}
+
 message InitialPayload {
   optional bool jumphost = 1 [default = false];
   repeated PortForwardSourceRequest reversetunnels = 2;
   map<string, string> environmentvariables = 3;
+  optional FlowControlMode flow_control_mode = 4 [default = FLOW_CONTROL_NONE];
 }
 
 message InitialResponse {
@@ -75,6 +90,9 @@ message ConfigParams {
 message TermInit {
   repeated string environmentnames = 1;
   repeated string environmentvalues = 2;
+  // Tells etterminal whether the attached client opted into flow control,
+  // so it can shrink the kernel buffer on its socket to the server.
+  optional FlowControlMode flow_control_mode = 3 [default = FLOW_CONTROL_NONE];
 }
 
 message TerminalUserInfo {
--- a/src/base/Headers.hpp
+++ b/src/base/Headers.hpp
@@ -367,6 +367,12 @@ inline bool waitOnSocketData(int fd) {
     if (errno == EINTR) {
       // Interrupted by the signal, the caller will retry.
       return false;
+    } else if (errno == EBADF || errno == EINVAL) {
+      // The fd was closed under us (e.g. peer disconnected). It will never
+      // become readable, and callers use this helper in retry loops, so
+      // returning false would spin at 100% CPU. Throw so the loop treats
+      // it as socket death.
+      throw std::runtime_error("Socket closed while waiting for data");
     } else {
       FATAL_FAIL(selectResult);
     }
@@ -389,8 +395,12 @@ inline bool isSocketWritable(int fd) {
   tv.tv_usec = 0;
   const int selectResult = select(fd + 1, NULL, &fdset, NULL, &tv);
   if (selectResult < 0) {
-    if (errno == EINTR) {
-      // Interrupted by the signal, the caller will retry.
+    if (errno == EINTR || errno == EBADF || errno == EINVAL) {
+      // EINTR: interrupted by a signal. EBADF/EINVAL: the fd was closed
+      // under us (e.g. client disconnected). Returning false is safe in
+      // both cases: callers use this as a non-blocking "can I write more?"
+      // poll and stop on false; a dead socket is then surfaced by the next
+      // read/write on the connection.
       return false;
     } else {
       FATAL_FAIL(selectResult);
--- a/src/base/PipeSocketHandler.cpp
+++ b/src/base/PipeSocketHandler.cpp
@@ -144,4 +144,18 @@ void PipeSocketHandler::stopListening(const SocketEndpoint& endpoint) {
   FATAL_FAIL(::close(sockFd));
 #endif
 }
+
+void PipeSocketHandler::minimizeKernelBuffering(int fd) {
+#ifndef WIN32
+  // Bound the kernel buffer on this unix socket. Terminal output waits
+  // here when the server applies backpressure (stops reading); the default
+  // (~200KB on Linux) is seconds of stale output on a slow link. 64KB does
+  // not limit throughput on a local socket.
+  int sndbuf = 64 * 1024;
+  if (setsockopt(fd, SOL_SOCKET, SO_SNDBUF, (char*)&sndbuf, sizeof(sndbuf)) <
+      0) {
+    LOG(WARNING) << "Failed to set SO_SNDBUF: " << strerror(errno);
+  }
+#endif
+}
 }  // namespace et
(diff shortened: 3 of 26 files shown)


MisterTea:
Cool!  I'm glad you had a chance to look into this!

Question: the way I typically use ET is I kick off a training job and then close my laptop. Will this stall the training process?  Do we need logic that disables flow control when there's no client connected?

bmaurer:
I've reworked this PR substantially to address the concern raised here and in #631 — that always-on backpressure would cause the "process blocks while disconnected" problem.

**The default is now byte-for-byte the current behavior.** Flow control only activates when a client explicitly passes `--flow-control backpressure` or `--flow-control discard`; the new proto field isn't serialized when unset, so old/new clients and servers interoperate unchanged. Nobody is affected unless they opt in.

The two opt-in modes map to the two camps in the issue thread:

- `backpressure` keeps today's lossless contract (slow client ⇒ process pauses, like plain ssh — safe for `tmux -CC`), but bounds the queue so Ctrl-C takes ~3s instead of a minute.
- `discard` is for the "never stall my job" use case: the process runs at full speed even through a 60s disconnect, and the display stays ~1s behind real time; the cost is that output exceeding the link rate is missing from scrollback.

The other significant finding while measuring this: application-level buffering alone barely helps, because the real queue lives in the kernel's autotuned TCP send buffer (up to 20MB on my test host) where nothing can be gated or dropped. That's why opted-in sessions set `TCP_NOTSENT_LOWAT` (32KB) — it only limits the *unsent* queue, not in-flight data, so fat-link throughput is preserved. Before/after on a 100KB/s link: Ctrl-C 56–116s → 2.1s, display lag unbounded → ~1.2s. Full methodology and numbers are in `test/e2e/REPORT.md`, and the harness (`test/e2e/do_scenario.sh`) reproduces every row of the table, including the disconnect scenarios.

The branch is now a 3-commit stack (standalone EBADF busy-loop fix → feature → E2E harness); each commit builds and passes the test suite independently.

bmaurer:
Rebased onto current master (v7.0.0) and re-validated everything.

The rebase was interesting because v7.0.0 independently landed adjacent work: disconnect buffering (#731, #762 — the connection now buffers up to 64MB while disconnected so processes don't freeze immediately), the PTY-input deadlock fix (#765), console output coalescing (#742), and port-forward fds in select() (#736). This PR now composes with all of that: the `none` path is exactly the new v7 code (including the `canBufferWrite` gating), coalescing is preserved in the opt-in client path, and the opt-in drain loops reuse the new `isSocketWritable()` helper (extended to tolerate a concurrently-closed fd instead of fatal-failing).

Re-measured on the v7.0.0 base, same harness (100KB/s link, 2.7MB/s producer):

| mode | display lag @30s | Ctrl-C → prompt |
|---|---|---|
| `none` (v7 baseline, unchanged) | ~32s, growing | 115.8s |
| `backpressure` (opt-in) | ~2s, bounded | 2.7s |
| `discard` (opt-in) | ~1.1s, bounded | 2.1s |

The v7 disconnect work doesn't change the connected-saturated-link behavior (writes still land in the autotuned multi-MB kernel send buffer), so #631 still reproduces on the baseline — the kernel-queue tuning here is what bounds it. Full test suite passes (130 tests) and the disconnect scenarios were re-run: in discard mode the producer runs at full speed through a 60s outage and the display recovers to ~1.1s of lag within one sample of reconnecting.

xz-dev:
We hit this failure mode in production and carried out bounded validation of this PR.

Under fast terminal output, ET without opt-in flow control can enter more than ordinary display lag: stale output grows into a multi-megabyte queue, recovery cannot drain it before `Socket Timeout`, the connection resets, and the next attempt starts over. The remote command, tmux session, and `etserver` may all remain alive while the client looks permanently frozen. Related reports:

- #631 — the unbounded flow-control/latency problem addressed by this PR
- #787 — repeated failed recovery attempts and retained recovery state
- #782 — reconnect/recovery state is not visible, so a recoverable or repeating stall presents as a frozen terminal

Our matched v7 baseline reproduced the loop: default `none` reached roughly 2.6–8.9 MB sampled Send-Q in controlled runs and failed recovery after the timeout. With explicit `--flow-control backpressure`, the producer stalled while congested, resumed after recovery, and display lag returned to roughly 1–2 seconds. `discard` also recovered, while intentionally dropping stale output. A separate real-path canary used an exact ~10 second outage, observed producer stall/resume, completed recovery with no `Socket Timeout`, and caught display lag back up below 2 seconds.

So for high-output remote terminals, this patch is not only a responsiveness improvement; it prevents a practical recovery deadloop that can otherwise continue while new output extends the backlog.

I prepared a focused follow-up PR directly against Ben's `backpressure-flow-control` branch: https://github.com/bmaurer/EternalTerminal/pull/1

That follow-up keeps Ben's existing three commits/authorship, merges current upstream master, resolves the sole #788 overlap, and repairs an existing E2E cleanup bug where fixed-port/process-pattern cleanup could kill unrelated processes. Validation: build passed; 159/159 tests passed with 2 privilege-dependent skips; formatting, occupied-port safety, tmux-session smoke, and PID-reuse cleanup checks passed.

MisterTea:
Closing in favor of #804.

Always-on `--flow-control backpressure` would stall a remote process after ~64KB while the client is disconnected, which fights ET's core use case (kick off a job, close the laptop, reconnect days later). Continuous `discard` avoids that stall but drops scrollback for the whole session.

#804 keeps lossless behavior by default. Unsent output is staged in a small userspace queue (with `TCP_NOTSENT_LOWAT` so it is not trapped in the kernel send buffer). Ctrl+C / Ctrl+Z / Ctrl+\ drop that queue only if it is already large (≥64KB), so the prompt is not stuck behind megabytes of stdout. While disconnected, PTY reads still use the existing 64MB reconnect buffer.

Thank you for the investigation on #631 — the kernel-queue finding (`TCP_NOTSENT_LOWAT`) is what makes a targeted flush possible.

PR #731, opened 2026-02-04
fix: buffer writes during disconnect to prevent programs in terminal from freezing

ejc3:
## Summary

When a client disconnects, `writePacket()` would loop forever waiting for the socket, causing the entire pipe chain to back up. This freezes any program writing to the terminal (e.g. Claude Code would no longer make progress).

This fix buffers data during disconnect (up to 4MB) before entering the blocking loop. On reconnect, buffered data is replayed via ET's existing recovery mechanism.

## Problem

The root cause was in `BackedWriter::write()`:
1. When `socketFd < 0`, it returned `SKIPPED` immediately **without buffering**
2. `Connection::write()` returned `false` on `SKIPPED`
3. `writePacket()` looped forever with 100ms sleeps waiting for reconnect
4. This blocked the terminal output pipe, filling the PTY buffer
5. Programs writing to stdout would block on write() and freeze

## Solution

- Buffer packets **before** checking socket state
- Add `BUFFERED_ONLY` return state for successful buffering without socket
- Add `DISCONNECT_BUFFER_BYTES` constant (4MB) to control when to start blocking
- Keep `MAX_BACKUP_BYTES` at 64MB for normal connected operation

## Behavior

| State | Action |
|-------|--------|
| Connected | Buffer + send (trim old at 64MB) |
| Disconnected, buffer < 4MB | Buffer data, return success (no blocking) |
| Disconnected, buffer >= 4MB | Return SKIPPED → writePacket() blocks until reconnect |

The blocking loop still exists but now only triggers after 4MB is buffered. This preserves data for replay on reconnect while still providing backpressure for extended outages.

## Changes

- `BackedWriter.hpp`: Add `BUFFERED_ONLY` enum, add `DISCONNECT_BUFFER_BYTES` constant
- `BackedWriter.cpp`: Buffer before socket check, respect disconnect limit
- `Connection.cpp`: Handle `BUFFERED_ONLY` as success
- `BackedIOTest.cpp`: Update existing test, add new disconnect buffer test

## Test plan

- [x] All 111 existing tests pass (942 assertions)
- [x] New unit test fills 4MB buffer and verifies SKIPPED is returned
- [x] Manual testing: disconnect/reconnect with active terminal output

## Evidence from production logs

After deploying this fix, the server logs show successful buffering during disconnects:

```
2026-02-04T06:30:23,822 Write buffered (no socket)
2026-02-04T06:30:23,932 Write buffered (no socket)
2026-02-04T06:00:23,821 Write buffered (no socket)
2026-02-04T06:00:23,934 Write buffered (no socket)
...
```

These writes completed successfully (returned `BUFFERED_ONLY`) instead of blocking. The terminal remained responsive during the disconnect, and output was replayed on reconnect.

### DIFF
--- a/src/base/BackedWriter.cpp
+++ b/src/base/BackedWriter.cpp
@@ -1,8 +1,8 @@
 #include "BackedWriter.hpp"
 
 namespace et {
-BackedWriter::BackedWriter(std::shared_ptr<SocketHandler> socketHandler_,  //
-                           std::shared_ptr<CryptoHandler> cryptoHandler_,  //
+BackedWriter::BackedWriter(std::shared_ptr<SocketHandler> socketHandler_,
+                           std::shared_ptr<CryptoHandler> cryptoHandler_,
                            int socketFd_)
     : socketHandler(socketHandler_),
       cryptoHandler(cryptoHandler_),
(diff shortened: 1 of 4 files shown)


MisterTea:
Thanks!

PR #736, opened 2026-03-05
Add port forward FDs to select() to reduce forwarding latency

splhack:
Port forwarding sockets were not included in the select() fd_set as the TODO comment, causing data to wait for the polling timeout (10ms client, 100ms server) before being read. Add them so select() returns immediately when forwarded data arrives.

https://github.com/MisterTea/EternalTerminal/blob/90b10d5f99be322d2ad9deabc4b86aa36a5f6894/src/terminal/TerminalClient.cpp#L207


splhack:
added unit tests.

MisterTea:
Thanks!!

PR #742, opened 2026-03-31
Coalesce console output

scramsby:
The ET client wrote each received TERMINAL_BUFFER packet to stdout individually. When the server sent a burst of output (e.g. a tmux session switch redrawing multiple panes), each packet became a separate write() to the local terminal, causing the terminal emulator to render intermediate states (screen-clear in one frame, repaint in the next) — producing visible flicker.
Fix: accumulate all TERMINAL_BUFFER payloads received in a single select() wake into one string, then write once after draining the packet queue.

To test:
Connect to Linux machine from Windows machine using ET.
Switch between tmux sessions with sidebar panes — verify no flicker and no partial screen erasure.


MisterTea:
Thanks!

PR #746, opened 2026-05-02
fix: pass long instead of int for SSH_OPTIONS_TIMEOUT (#737)

### LINKED ISSUE #737: ARM64: 'invalid error' on connect due to int/long type mismatch in SSH config parser
## Description

When connecting via `et` on ARM64 macOS, the message `invalid error` is printed to stdout before the connection succeeds. The connection works fine — the message is cosmetic but confusing.

## Environment

- **Client**: et 6.2.11, macOS (Apple Silicon M4, ARM64)
- **Server**: etserver 6.2.10, Ubuntu 24.04 (x86_64)
- **SSH config trigger**: `ConnectTimeout 10` in `Host *` block of `~/.ssh/config`

## Root Cause

In `src/terminal/ParseConfigFile.hpp`, the `SOC_TIMEOUT` case in `ssh_config_parse_line()` stores the parsed timeout value in a local `int i` (4 bytes) and passes `&i` to `ssh_options_set()`.

However, `ssh_options_set()` for `SSH_OPTIONS_TIMEOUT` casts the pointer to `long *` and dereferences it — reading 8 bytes on ARM64 (where `sizeof(long) == 8`). The extra 4 bytes are uninitialized stack data. When the high bit is set, the value appears negative, triggering:

```c
long *x = (long *)value;  // reads 8 bytes from a 4-byte int
if (*x < 0) {
    CLOG(INFO, "stdout") << "invalid error" << endl;
    return -1;
}
```

The `"stdout"` logger uses format `"%msg"`, so it prints bare `invalid error` with no context.

## Steps to Reproduce

1. Add `ConnectTimeout 10` to the `Host *` block in `~/.ssh/config`
2. Connect from ARM64 macOS: `et user@host`
3. Observe `invalid error` printed before successful connection

## Impact

- Cosmetically confusing — no indication of what's "invalid"
- Functionally harmless — connection succeeds because the return value of `ssh_options_set` is ignored for this case
- Only affects **64-bit platforms where `sizeof(long) != sizeof(int)`** (ARM64 macOS, x86_64 Linux)

## Suggested Fix

Change the type from `long *` to `int *` in the `SSH_OPTIONS_TIMEOUT` cast (and audit similar casts in `ssh_options_set` for the same pattern). Alternatively, change the local variable in `ssh_config_parse_line()` from `int i` to `long i`.

vzd3v:
## Summary

Fixes #737 — the spurious `invalid error` message printed to stdout on connect when `~/.ssh/config` contains `ConnectTimeout`.

## Root cause

`ssh_config_parse_line()` (`src/terminal/ParseConfigFile.hpp`) shares a single `int i` across cases. For the `SOC_TIMEOUT` case it does:

```cpp
i = ssh_config_get_int(&s, -1);
ssh_options_set(options, SSH_OPTIONS_TIMEOUT, &i);  // passes int*
```

But `SSH_OPTIONS_TIMEOUT` is documented as `long` (line ~656) and the handler in `ssh_options_set()` casts the pointer accordingly:

```cpp
case SSH_OPTIONS_TIMEOUT:
  ...
  long *x = (long *)value;   // reads 8 bytes
  if (*x < 0) {
    CLOG(INFO, "stdout") << "invalid error" << endl;
    return -1;
  }
```

On LP64 platforms (arm64 macOS, x86_64 Linux), the handler reads 8 bytes from a 4-byte stack slot, picking up uninitialized upper bits. Whenever those bits happen to have the sign bit set, the handler treats the value as negative and prints `invalid error` before returning. The connection still succeeds because the return value of `ssh_options_set` is not checked here, so the symptom is purely cosmetic — but confusing.

This only triggers when `ConnectTimeout` actually appears in the user's ssh config (it's the only keyword mapped to `SOC_TIMEOUT`).

## Fix

Use a local `long timeout` for this case so the type matches the consumer. Scoped with braces to keep the new local out of other cases. No change to the handler API or to any other case.

## Test plan

- [x] Builds cleanly (single-line type change in one case)
- [x] Reproduce: add `ConnectTimeout 10` under `Host *` in `~/.ssh/config`, run `et user@host` on arm64 macOS — `invalid error` no longer appears
- [x] Connection still works as before; remove the line and behavior is identical

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/terminal/ParseConfigFile.hpp
+++ b/src/terminal/ParseConfigFile.hpp
@@ -1433,12 +1433,18 @@ static int ssh_config_parse_line(const char *targethost,
         SAFE_FREE(b);
       }
       break;
-    case SOC_TIMEOUT:
-      i = ssh_config_get_int(&s, -1);
-      if (i >= 0 && *parsing) {
-        ssh_options_set(options, SSH_OPTIONS_TIMEOUT, &i);
+    case SOC_TIMEOUT: {
+      // SSH_OPTIONS_TIMEOUT is documented as `long` and the handler casts
+      // value to `long *`. Passing `int *` made the handler read 8 bytes
+      // from a 4-byte slot on LP64 (arm64 macOS, x86_64 Linux), so
+      // uninitialized upper bits could flip the sign and trigger a bogus
+      // "invalid error" message. See issue #737.
+      long timeout = ssh_config_get_int(&s, -1);
+      if (timeout >= 0 && *parsing) {
+        ssh_options_set(options, SSH_OPTIONS_TIMEOUT, &timeout);
       }
       break;
+    }
     case SOC_STRICTHOSTKEYCHECK:
       i = ssh_config_get_yesno(&s, -1);
       if (i >= 0 && *parsing) {


MisterTea (approved):
Thanks!

PR #753, opened 2026-05-22
[Windows] Reassert autowrap after console setup

scramsby:
Description:
Windows ET setup changes console output modes to enable VT processing and DISABLE_NEWLINE_AUTO_RETURN. On affected Windows terminal hosts, long readline input can then repaint over the same visual row instead of wrapping, even though a DECRQM query still reports DECAWM enabled. Preserve the existing console mode bits and explicitly include ENABLE_WRAP_AT_EOL_OUTPUT so setup does not unnecessarily clear normal wrapping-related console state. That alone was tested and was not sufficient to fix the regression; the empirically required step is re-emitting DEC autowrap enable after the output mode change. DISABLE_NEWLINE_AUTO_RETURN remains intentional because removing it would regress full-screen terminal apps such as tmux, where it suppresses incorrect scrolling when content reaches the right edge of the terminal window.

Test plan:
Built the Windows et target locally; verified a raw ET session from Windows wraps long command input correctly after reconnecting; verified tmux full-screen scrolling still works without duplicated status elements from improper terminal-edge scrolling

### DIFF
--- a/src/terminal/PseudoTerminalConsole.hpp
+++ b/src/terminal/PseudoTerminalConsole.hpp
@@ -5,6 +5,10 @@
 #include "ETerminal.pb.h"
 #include "RawSocketUtils.hpp"
 
+#ifdef WIN32
+#include <iostream>
+#endif
+
 namespace et {
 /**
  * @brief Configures the local console into raw mode and exposes terminal info.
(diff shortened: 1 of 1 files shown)


MisterTea:
Thanks!!

PR #762, opened 2026-06-10
Keep sessions alive while the client is disconnected

davidblurton:
Sessions die when a client goes away (e.g. laptop lid closed) while the shell keeps producing output. The failure chain, observed in production logs:

1. The disconnect buffer (4MB) fills with pending terminal output.
2. The server's session thread blocks forever in `writePacket` (`Waiting to write...`).
3. The etterminal's router socket stops draining, and macOS eventually fails the etterminal's blocked write with ETIMEDOUT (`Failed a call to writeAll: Operation timed out`).
4. The etterminal exits, and the reconnecting client gets `INVALID_KEY` — the session is gone.

This fixes each link in the chain by propagating backpressure to the shell's pty instead of blocking or dying:

- **TerminalServer** only reads from the terminal router when the client connection can absorb the data.
- **UserTerminalHandler** only reads the pty when the router socket is writable, and **SocketHandler** retries ETIMEDOUT on writes when the caller didn't request a timeout (macOS raises it on unix sockets whose peer stops draining, even though the connection is intact).
- **BackedWriter** counts the disconnect limit against bytes buffered since the disconnect (not total backup size, which can already be at its cap with delivered data), and raises it from 4MB to 64MB.
- **Connection** treats wake-from-sleep errnos (ENETDOWN, ENETUNREACH, ENETRESET, EHOSTDOWN, EPROTOTYPE) as skippable, and severs and reconnects on unexpected errnos instead of throwing (which kills the session) or aborting (which on a server kills every session).

Tested with a sleep/wake harness: a client SIGSTOP'd behind a killed TCP proxy with ~1MB/s of terminal output during a 45 second outage now recovers and replays all output. Unit tests added for the disconnect-buffer accounting, the ETIMEDOUT retry, and severing on unexpected read errnos.

### DIFF
--- a/src/base/BackedWriter.cpp
+++ b/src/base/BackedWriter.cpp
@@ -8,15 +8,17 @@ BackedWriter::BackedWriter(std::shared_ptr<SocketHandler> socketHandler_,
       cryptoHandler(cryptoHandler_),
       socketFd(socketFd_),
       backupSize(0),
+      disconnectedBytes(0),
       sequenceNumber(0) {}
 
 BackedWriterWriteState BackedWriter::write(Packet packet) {
   // If recover started, Wait until finished
   lock_guard<std::mutex> guard(recoverMutex);
 
-  // If no socket and buffer exceeds disconnect limit, signal caller to wait
-  // This prevents blocking during normal disconnects while preserving data
-  if (socketFd < 0 && backupSize + packet.length() > DISCONNECT_BUFFER_BYTES) {
+  // If no socket and the data buffered since the disconnect exceeds the
+  // limit, signal caller to wait
+  if (socketFd < 0 &&
+      disconnectedBytes + packet.length() > DISCONNECT_BUFFER_BYTES) {
     return BackedWriterWriteState::SKIPPED;
   }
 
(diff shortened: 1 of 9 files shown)


MisterTea:
"only reads the pty when the socket is available"

How does this affect long running tasks?  If I have an AI training job that logs to stderr, will the job block while the client is disconnected?

davidblurton:
Not until 64MB of output. While the client is disconnected the server keeps draining the pty into the disconnect buffer `DISCONNECT_BUFFER_BYTES`, raised from 4MB to 64MB. This buffer now counts bytes-since-disconnect rather than total backup. The job only feels back-pressure once 64MB accumulates after disconnecting. The server stops reading the pty, kernel pty buffer fills, then the job's next stderr write blocks. It resumes when the client reconnects and drains. CPU-only work between writes never blocks.

Today the same scenario stalls much sooner: the disconnect limit (4MB, counted against the total backup buffer, so often less) trips, the session thread spins in writePacket and the et terminal stops being drained. On Linux the job blocks, on macOS the et terminal's blocked socket write fails with `ETIMEDOUT`, exits, and the reconnecting client gets `INVALID_KEY` — session and pty (`SIGHUP`) gone.

I'm currently using this for a similar use-case - detaching and reattaching to remote Claude code session.

MisterTea:
Makes sense. Thanks for the PR!

Issue #682, opened 2025-02-25
pasting into vim hangs on large pastes


jhaar:
I've been using "et" for a couple of months now and have noticed that if I paste something "large", the paste starts and then hangs *HARD* (right now it happened with 240K CSV data I was copying from a Chrome Sheet to vim). Weirdly it only pastes a few lines (out of hundreds/thousands) before hanging, and I cannot even recover the "et" session - I end up having to kill the whole terminal session and start again.

BTW I don't actually think it's got anything to do with vim - it's just that's really the only thing I would be pasting large chunks of data into...

This is from my Ubuntu-24.04 laptop (apt install of et-6.2.9-noble1) to a Fedora-40 server (yum install of et-6.2.8-1.fc40.x86_64)

Any ideas? Thanks!

Kronuz:
Confirming the diagnosis, and I think that this is a server-side deadlock in the per-session terminal pump.

`UserTerminalHandler::runUserTerminal` is a single-threaded `select()` loop that writes client input to the pty master with a blocking `RawSocketUtils::writeAll()`. On a large paste:

1. ET writes the paste to the master.
2. The shell echoes it back, filling the pty's output buffer.
3. With the output buffer full, the shell stops reading its input.
4. The blocking writeAll() to the master never returns.
5. So the same loop never gets back to draining output. Hard deadlock, and the session is unrecoverable, exactly as you describe.

The fix is to make the master non-blocking at creation and rework the pump so it never blocks. I have this running in a fork with a regression test that drives a real pty through the real handler. I'll open a PR.

cc: @MisterTea

MisterTea:
Closing this as fixed. PR #765 directly fixes the blocking PTY input-write deadlock described by the reporter.

PR #765, opened 2026-06-18
Fix input-write deadlock in UserTerminalHandler on large input bursts

Kronuz:
## The bug

The per-session terminal pump, `UserTerminalHandler::runUserTerminal`, is a single-threaded `select` loop that wrote client input to the pty master with a **blocking** `RawSocketUtils::writeAll()`. A burst of input larger than ~one pty buffer deadlocks the session:

1. The input echoes back and fills the pty's output buffer.
2. With its output buffer full, the shell stalls and stops reading input.
3. The blocking write to the master never drains, so it never returns.
4. Stuck in that write, the loop also stops draining output -- the very thing that would unstick the shell.

The whole session wedges. It's reproducible by pasting a multi-KB heredoc: ~1KB on macOS, ~8-10KB on Linux (where one pty buffer is larger).

Related is [issue 682](https://github.com/MisterTea/EternalTerminal/issues/682).

## The fix

The read path already tolerated `EAGAIN` -- it was written expecting a non-blocking master -- but nothing ever made the master non-blocking. So:

- **Set `O_NONBLOCK` where the fd is created**, in `UserTerminal::setup` (`PseudoUserTerminal` and the test terminals). This is now part of that interface's contract.
- **Rework the pump to never block:** buffer pending input, drain it to the master through the `select` write set, and keep reading output every iteration.
- **Add backpressure:** when the input buffer fills, stop reading more from the router so backpressure reaches the client -- symmetric to the existing output gating on `routerWritable`.

## Testing

A new regression test, `TerminalTest "LargeInputNoDeadlock"`, drives a real pty (`cat` in raw mode) through the real handler. It first does a small warmup round-trip to confirm the full client -> pty -> client echo path is live, then pushes an 8KB payload (well past one pty buffer) and requires every byte to echo back within a generous deadline. The old code deadlocks and the read never completes, so the test **times out and fails; with the fix it round-trips and passes.** The warmup keeps the test reliable on a slow build (a sanitizer run under `ctest --parallel`), where it would otherwise start typing before the pipeline was wired; it is far smaller than one pty buffer, so it round-trips even on the buggy code and cannot mask the regression. The existing tests never caught this because they all use `FakeUserTerminal` (a socket pair), which has no pty buffer semantics.

The full `et-test` suite is green, and the fix has been validated live: a 56KB one-shot input that previously wedged the session now round-trips byte-exact against a real `etserver`.

## Why it's low-risk

**Server-side only -- no protocol, client, or proto changes.** The change is confined to the terminal pump and the `UserTerminal::setup` contract; the wire format and everything around it are untouched. A non-blocking master is what the read path already assumed, so this makes the code match its own expectations.

## Asking for guidance

Setting `O_NONBLOCK` in `setup` (rather than inside the pump) made the contract explicit and kept the fake terminals honest, but I'm happy to move it if you'd rather. Is the approach right?

---

*This PR was prepared with AI assistance; I've reviewed and stand behind it.*

### DIFF
--- a/src/terminal/PseudoUserTerminal.hpp
+++ b/src/terminal/PseudoUserTerminal.hpp
@@ -1,6 +1,7 @@
 #ifndef __PSUEDO_USER_TERMINAL_HPP__
 #define __PSUEDO_USER_TERMINAL_HPP__
 
+#include <fcntl.h>
 #include <stdlib.h>
 #include <sys/types.h>
 #include <sys/wait.h>
@@ -53,6 +54,16 @@ class PseudoUserTerminal : public UserTerminal {
       utempter_add_record(masterFd, buf);
     }
 #endif
+
+    // The handler polls this fd with select() and does non-blocking reads and
+    // writes, so make the master non-blocking here, where it is created.  If it
+    // stayed blocking, a large input burst would block the handler's single
+    // write() call and deadlock against the shell's echo (see
+    // UserTerminal::setup and UserTerminalHandler::runUserTerminal).
+    int flags = fcntl(masterFd, F_GETFL, 0);
+    if (flags != -1) {
+      fcntl(masterFd, F_SETFL, flags | O_NONBLOCK);
+    }
     return masterFd;
   }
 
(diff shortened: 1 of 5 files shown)


MisterTea:
thanks!

Issue #744, opened 2026-04-26
Installation fails on Ubuntu Desktop 26.04 LTS (Resolute Raccoon)


grigutis:
Adding the PPA fails since there is no `resolute` directory:

```
$ sudo add-apt-repository ppa:jgmath2000/et
PPA publishes dbgsym, you may need to include 'main/debug' component
Repository: 'Types: deb
URIs: https://ppa.launchpadcontent.net/jgmath2000/et/ubuntu/
Suites: resolute
Components: main
'
Description:
Eternal Terminal (ET) is a remote shell that automatically reconnects without interrupting the session. ET survives IP roaming and other connection loss events.
More info: https://launchpad.net/~jgmath2000/+archive/ubuntu/et
Adding repository.
Press [ENTER] to continue or Ctrl-c to cancel.
Found existing deb entry in /etc/apt/sources.list.d/jgmath2000-ubuntu-et-resolute.sources
Hit:1 http://security.ubuntu.com/ubuntu resolute-security InRelease
Hit:2 http://us.archive.ubuntu.com/ubuntu resolute InRelease
Hit:3 http://us.archive.ubuntu.com/ubuntu resolute-updates InRelease
Hit:4 http://us.archive.ubuntu.com/ubuntu resolute-backports InRelease
Ign:5 https://ppa.launchpadcontent.net/jgmath2000/et/ubuntu resolute InRelease
Err:6 https://ppa.launchpadcontent.net/jgmath2000/et/ubuntu resolute Release
  404  Not Found [IP: 185.125.190.80 443]
Reading package lists... Done
E: The repository 'https://ppa.launchpadcontent.net/jgmath2000/et/ubuntu resolute Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

BobDu:
Same issue on Ubuntu 26.04 LTS (resolute) — the PPA has no resolute suite yet (latest published is questing / 25.10), so add-apt-repository ppa:jgmath2000/et 404s.

You can still install it through apt — just point the PPA at the nearest release (questing / 25.10) instead of resolute. add-apt-repository still imports the signing key, and the 25.10 build runs fine on 26.04
(dependency sonames are unchanged):
```
# adds the PPA + imports its signing key (the 'update' step 404s here — expected)
sudo add-apt-repository -y ppa:jgmath2000/et

[...]

Issue #581, opened 2023-06-29
build error: "/usr/bin/ld: cannot find -lstdc++fs: No such file or directory"


oliverkurth:
I am trying to build an RPM for ET in Photon 5.0, but I am running into this error:
```
[ 84%] Linking CXX executable htm
/usr/bin/cmake -E cmake_link_script CMakeFiles/htm.dir/link.txt --verbose=1
/usr/bin/c++ -O2 -g -DUSE_SENTRY -DELPP_NO_DEFAULT_LOG_FILE -DELPP_FEATURE_CRASH_LOG -DELPP_THREAD_SAFE -DELPP_STRICT_PERMISSIONS -DSENTRY_BUILD_STATIC -g -ggdb3 -DWITH_UTEMPTER -rdynamic CMakeFiles/htm.dir/src/htm/HtmClientMain.cpp.o -o htm  libHtmCommon.a libet-lib.a -lprotobuf-lite -lsodium -lutempter external_imported/sentry-native/libsentry.a /usr/lib/libssl.so /usr/lib/libz.so -lutil -lresolv -latomic -lstdc++fs /usr/lib/libcurl.so -Wl,-E,--build-id=sha1 -ldl -lrt external_imported/sentry-native/external/libbreakpad_client.a /usr/lib/libcrypto.so 
/usr/bin/ld: cannot find -lstdc++fs: No such file or directory
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/htm.dir/build.make:120: htm] Error 1
make[2]: Leaving directory '/usr/src/photon/BUILD/EternalTerminal-et-v6.2.1/x86_64-vmware-linux'
make[1]: *** [CMakeFiles/Makefile2:707: CMakeFiles/htm.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

I have the following build requires in the spec file:
```
BuildRequires:  boost-devel
BuildRequires:  cmake
BuildRequires:  gflags-devel
BuildRequires:  curl-devel
BuildRequires:  libsodium-devel
BuildRequires:  libstdc++-devel
BuildRequires:  libutempter-devel
BuildRequires:  ncurses-devel
BuildRequires:  openssl-devel
BuildRequires:  protobuf-devel
BuildRequires:  protobuf-c-static
BuildRequires:  protobuf-c-devel
BuildRequires:  systemd
BuildRequires:  zlib-devel
```

[...]

PR #774, opened 2026-07-12
Fix GCC 15 build failure — Ubuntu 26.04 (Resolute) PPA

chenlichao:
## Problem

The PPA build for Ubuntu 26.04 (Resolute) fails on Launchpad ([build #33370536](https://launchpad.net/~jgmath2000/+archive/ubuntu/et/+build/33370536) — FAILEDTOBUILD). Tracked in #744.

Ubuntu 26.04 ships **GCC 15**, which introduces two breaking changes:

### 1. `-lstdc++fs` no longer exists (linker error)

GCC 15 completely removed the separate `stdc++fs` library — filesystem support was merged into `libstdc++` proper since GCC 9. The CMakeLists.txt unconditionally links `stdc++fs` on Linux, causing:

```
/usr/bin/ld: cannot find -lstdc++fs: No such file or directory
```

Same class of issue as #581.

### 2. Implicit `<cstdint>` includes removed (compiler errors)

GCC 15 removed transitive header inclusions. Source files using `uint64_t`/`uint32_t` without explicit `#include <cstdint>` now fail:

```
error: 'uint64_t' does not name a type
```

## Fix

**CMakeLists.txt** — Replace unconditional `stdc++fs` linking with `check_library_exists()` so it only links when the library is present (GCC ≤ 8). Backwards compatible.

**8 source files** — Added `#include <cstdint>` to files using fixed-width integer types without the explicit include:
- `src/htm/MultiplexerState.cpp`
- `src/htm/HtmServer.cpp`
- `src/terminal/UserTerminalHandler.cpp`
- `src/terminal/TerminalServer.cpp`
- `src/terminal/TerminalClient.cpp`
- `src/terminal/forwarding/PortForwardHandler.cpp`
- `src/base/BackedWriter.cpp`
- `src/base/UnixSocketHandler.cpp`

## Related

- #744 — Installation fails on Ubuntu 26.04
- #581 — `-lstdc++fs` linker error (same root cause, different distro)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -326,8 +326,16 @@ else()
     util
     resolv
     atomic
-    stdc++fs
   )
+  # stdc++fs was a separate library for <filesystem> in GCC 8 and earlier.
+  # Starting with GCC 9, filesystem support is part of libstdc++ proper, and
+  # the separate -lstdc++fs was removed entirely in GCC 15.
+  # Only link it if the compiler still ships it.
+  include(CheckLibraryExists)
+  check_library_exists(stdc++fs "" "" HAVE_STDCXXFS)
+  if(HAVE_STDCXXFS)
+    list(APPEND CORE_LIBRARIES stdc++fs)
+  endif()
 endif()
 
 IF(Unwind_FOUND)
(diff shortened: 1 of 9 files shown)


MisterTea:
Thanks!

PR #777, opened 2026-07-15
Add windows deployment workflows

hyec:
Add a workflow for publishing Windows release artifacts.

### DIFF
--- a/.github/workflows/deploy_windows_winget.yml
+++ b/.github/workflows/deploy_windows_winget.yml
@@ -1,6 +1,8 @@
 name: Deploy Windows Release Asset
 
 on:
+  release:
+    types: [published]
   workflow_dispatch:
     inputs:
       release_tag:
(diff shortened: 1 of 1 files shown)


hyec:
To provide full Windows package support through WinGet, I plan to complete the following steps:
1. Merge this PR and publish a new Windows release artifact.
2. Submit the WinGet manifest currently prepared in [hyec/winget-pkgs](https://github.com/hyec/winget-pkgs/tree/MisterTea.EternalTerminal-7.0.0-0e88bd98-7101-4cc7-995c-73245fb3ccef). This needs to be done after the first step because the manifest requires the SHA256 of the installer archive. In the meantime, the package identifier, description, and other manifest metadata can already be reviewed.
3. Add a WinGet manifest update workflow that submits a new PR to microsoft/winget-pkgs when a new Eternal Terminal version is released.


There is one additional point I would like to discuss: where should the Windows artifact be hosted?

My suggestion is to publish it as a GitHub Release asset, as implemented in this workflow. WinGet does not require the project to maintain a repository like an APT repository or Homebrew tap; it only needs a download URL for an installer archive.

However, uploading only a Windows artifact to every release may seem weird. We could also consider publishing the Debian, Ubuntu and macOS packages as release assets.

MisterTea:
Thanks!

PR #788, opened 2026-08-06
Keep the client alive when the console fd is not a tty

jshort:
ddf67a5 (#739) replaced FATAL_FAIL(rc) in TerminalClient::run()'s console read loop with explicit handling of every read() outcome. That correctly fixed the spurious ENOENT crash, but it also turned rc == 0 into an unconditional "clean session end":

    } else if (rc == 0) {
      LOG(INFO) << "Console EOF";
      break;
    }

That treats EOF as "the user closed an interactive session", which only holds when the console descriptor is actually a terminal. Often it is not.

Console::getFd() returns STDOUT_FILENO, so the loop reads *stdout*, not stdin. At a normal terminal, fds 0, 1 and 2 all refer to the same tty, so reading fd 1 works and the distinction never shows. Under nohup(1) it does: nohup redirects stdout to nohup.out while leaving the tty on stdin, so the client ends up reading a regular file. select() always reports a regular file ready, and the read always returns 0 -- the file is empty at startup, and the shared offset then tracks the end of the file as the client writes terminal output to it. The client therefore exits moments after connecting, before the session can be used:

    [INFO] Got command: echo hello
    [INFO] Console EOF
    [INFO] Shutting down connection

This breaks any workflow that backgrounds the client so it outlives the process that launched it: nohup et host &, service managers, and IDE remote-development integrations that hold port forwards open across editor restarts. A plain redirect is affected too -- et host --command '...' > out.txt opens stdout O_WRONLY, so read() returns EBADF and takes the other break added by #739.

Before #739 this was invisible: FATAL_FAIL(X) fires only on X == -1, so rc == 0 fell through silently and the session survived, busy-looping on a descriptor it could never read from.

Treat EOF and hard read errors as session-fatal only when the console descriptor is a tty. Otherwise stop selecting on it and keep running. Console input is impossible either way -- it always was under nohup -- but the session and its port forwards stay up. Terminal output is unaffected: it still goes to the same descriptor.

This completes #739 rather than reverting any of it. Every read() outcome is still handled explicitly and the errno-corruption crash stays fixed; a non-interactive console simply no longer ends the session.

Testing:

Adds NonTtyConsoleKeepsSessionAlive to test/integration_tests/TerminalTest.cpp. It drives a real TerminalClient with a FileBackedConsole whose descriptor is a regular file opened O_RDWR|O_APPEND -- the shape nohup installs -- and requires that run() has not returned once the connection is up, and that remote output still reaches that descriptor. Pre-fix, run() returns within milliseconds.

Verified manually on macOS: with a patched client, a remote session launched under nohup stays up and its port forwards carry traffic, where an unpatched build exited immediately after login.

### DIFF
--- a/src/terminal/TerminalClient.cpp
+++ b/src/terminal/TerminalClient.cpp
@@ -180,6 +180,13 @@ void TerminalClient::run(const string& command, const bool noexit) {
     CLOG(INFO, "stdout") << "ET running, feel free to background..." << endl;
   }
 
+  // Launchers such as nohup(1) replace stdout with a regular file while
+  // leaving the tty on stdin, so the descriptor Console exposes for input is
+  // readable-but-empty (or not readable at all). select() always reports it
+  // ready and every read yields EOF/EBADF, which must not be mistaken for the
+  // user closing an interactive session -- doing so would tear down a
+  // backgrounded client and its port forwards immediately.
+  bool consoleInputDisabled = false;
   while (!connection->isShuttingDown()) {
     {
       lock_guard<recursive_mutex> guard(shutdownMutex);
@@ -195,7 +202,7 @@ void TerminalClient::run(const string& command, const bool noexit) {
     FD_ZERO(&rfd);
     int maxfd = -1;
     int consoleFd = -1;
-    if (console) {
+    if (console && !consoleInputDisabled) {
       consoleFd = console->getFd();
       maxfd = consoleFd;
       FD_SET(consoleFd, &rfd);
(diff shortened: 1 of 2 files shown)


MisterTea:
@jshort should I cut a new release to get this in?

jshort:
@MisterTea we encountered this because, coincidentally and quite recently, an IDE integration we use that connects to hosts via ET uses nohup to setup the connection tunnels.  It was this use of no-hup with a v7.0.0 client that caused the issues that this PR addresses so I don't know if it is pervasive.

PR #790, opened 2026-08-12
Keep destination SSH options off the jumphost

sdmcclain:
## Summary

- stop replaying destination-scoped `--ssh-option` values on the independent jumphost relay connection
- preserve the jumphost user, port, address, alias, and IPv6 parsing added in #705
- add a regression test that records both SSH calls and verifies target options stay on the target bootstrap

## Rationale

Jumphost mode launches two SSH processes with different destinations and often different users, identities, host keys, and trust policies. Copying target-only options onto the direct jumphost call can both break valid ProxyJump setups and unintentionally weaken jumphost verification.

Jump-specific policy continues to resolve independently from the jumphost's SSH configuration.

## Testing

- source and test formatting checks pass
- Linux ASan, UBSan, MSan, TSan, jumphost system, and backwards-compatibility jobs pass
- Codecov reports all modified coverable lines exercised
- `git diff --check`; patch applies cleanly to both `master` and the `release`/v7.0.0 branch

PR #791, opened 2026-08-13
Allow callers to select an exact SSH config

sdmcclain:
## Summary

- Add --ssh-config <absolute-path|none> so callers can select one SSH configuration policy.
- Add --no-ssh-config as an alias for --ssh-config none.
- Use the selected policy for ET destination and jumphost resolution, and pass the same -F value to both SSH subprocess shapes.
- Preserve the existing ambient user/system configuration behavior when neither option is present.

## Motivation

ET currently parses user and system SSH configuration before starting SSH. A managed client that already owns its SSH policy cannot prevent ambient HostName, ProxyJump, LocalForward, or SetEnv directives from changing the target or session behavior.

An exact configuration root lets such clients supply a private, auditable SSH policy without modifying user dotfiles or inheriting unrelated settings.

## Safety and behavior

An explicit file must be absolute, readable, regular, non-symlinked, and use a conservative path alphabet (ASCII letters, digits, slash, dot, underscore, and hyphen). Metadata is checked before opening the file, so non-regular paths such as FIFOs fail without blocking. The alphabet restriction matters because OpenSSH constructs its implicit ProxyJump command through a shell and propagates -F without quoting its value.

The selected file is used by ET for destination and jumphost lookup and is passed to the destination bootstrap and direct jumphost SSH processes. OpenSSH propagates that -F setting to the implicit ProxyJump child. The none form skips ET config parsing and invokes SSH with -F none.

## Validation

- Ubuntu 22.04 arm64 no-vcpkg build of et and et-test passed.
- Focused [SshSetupHandler] suite passed: 69 assertions in 10 test cases.
- CLI probes passed for exact-file selection, none, relative paths, unsafe paths, missing files, symlinks, directories, FIFOs, and mutually exclusive options. The FIFO probe used a timeout and confirmed immediate status 1 rejection rather than an open/block.
- clang-format 18.1.8 dry-run passed on every changed C++ file.
- git diff --check passed.

## Compatibility with #790

This change has no semantic dependency on #790 and intentionally does not change the existing --ssh-option behavior. Its tests use no destination SSH options, so their expectations are valid before or after #790. As a textual compatibility check, the #790 patch applies cleanly on top of this branch with every hunk succeeding. Either PR can land first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,11 +231,13 @@ To isolate ET from ambient SSH configuration, pass an absolute path with
 `--ssh-config`. ET reads only that file for its own destination and jumphost
 lookup and passes the same file to every SSH process it starts, including the
 implicit ProxyJump connection. The file must be readable, regular, and not a
-symbolic link, and its absolute path may contain only ASCII letters, digits,
-`/`, `.`, `_`, and `-`; OpenSSH does not quote this path when constructing its
-implicit ProxyJump command. Pass `--ssh-config none` or the equivalent
-`--no-ssh-config` to disable both user and system SSH configuration entirely.
-The two options are mutually exclusive.
+symbolic link. The path must be absolute (`/path` on Unix; `C:\path`,
+`C:/path`, or a UNC path on Windows) and may contain only ASCII letters, digits,
+`/`, `.`, `_`, and `-` (plus `:` and `\` on Windows). Spaces are not allowed:
+OpenSSH does not quote this path when constructing its implicit ProxyJump
+command. Pass `--ssh-config none` or the equivalent `--no-ssh-config` to
+disable both user and system SSH configuration entirely. The two options are
+mutually exclusive.
 
 ## Building from Source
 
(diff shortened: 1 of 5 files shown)


MisterTea:
Rebased onto current `master` and added Windows support for `--ssh-config`.

`--ssh-config` now uses `std::filesystem::path::is_absolute()` instead of requiring a leading `/`, and on Windows the ProxyJump-safe alphabet also allows `:` and `\` so drive and UNC paths work. Spaces are still rejected (OpenSSH interpolates `-F` unquoted; Windows `CreateProcess` also concatenates argv without quoting).

MisterTea:
@copilot rebase this on latest master

MisterTea:
@sdmcclain copilot isn't available for cross repo PRs. Can you get PRCI working?

MisterTea:
Taking this over: Copilot cannot push to a cross-repo PR, and CI was stuck on a FreeBSD checkout failure after the last rebase.

Rebased onto current `master` and combined with #790 in https://github.com/MisterTea/EternalTerminal/pull/814. Closing this PR in favor of that one. Credit remains with @sdmcclain for the original change; the Windows path follow-up is included.

PR #793, opened 2026-08-20
Named sessions and etserver restart survival

epsalmond:
Sessions get a name, persist across client crashes and reboots, and survive `etserver` restarts.

I started this work because ubuntu was killing all my sessions at 3am doing updates. Then continued because (surprise!) there was a memory leak in node eating all the memory on my laptop. And an NFS bug in osx was rebooting me.

For the last month, having ~20 sessions open across tabs in two ghostty windows has been my primary workflow.

QoL improvements have happened along the way including being noisier when stealing a session from an active session. Further improvements could be detecting and cleaning up stale sessions.

## Client

- `et --name <n>` creates or reattaches a named session. Direct connections persist by default with a short generated name; `--no-persist` opts out.
- `et --list` shows saved sessions with their captured terminal title and last heartbeat. `--attach` reattaches, `--kill` terminates a named session on the server.
- Session files live in `~/.et/sessions/` with owner-only permissions, atomic no-clobber writes, and credential-safe SSH bootstrap diagnostics.
- A read-only recovery audit plus explicitly confirmed recovery import with private backups.

## Protocol

- Reset handshake so a fresh client process can reattach to a live server connection. Each reset epoch derives fresh crypto keys.
- Additive proto fields only. Protocol version is unchanged and the backwards-compatibility test passes.

## Server

- `etterminal` keeps the pty alive when the router drops and re-registers with bounded backoff, so `systemctl restart et` no longer kills sessions.
- A restart grace window answers reconnects with retry-later instead of a hard rejection. Reverse tunnels dropped by a restart are reported.
- Bounded reconnect bookkeeping, replaced-registration handling outside the router lock, and shared session teardown for fresh and resumed connections.

## Tests

- Session store, reset handshake, router restart, and named-session system tests. Suite is green under normal, ASan, and TSan builds, and rebased onto current master including the epoll and Windows parity changes.

I'm working on resolving these:

  High:
  - Reset recovery can be triggered by a known client ID without passkey proof, allowing disruption/data loss.

  Medium:

  - Windows router restart and --kill behavior are incomplete.
  - Restart during initial Unix setup is not recovered.
  - external/UniversalStacktrace moves backward to an ancestor commit.
  - --name can silently ignore newly requested forwarding options.
  - Fixed /tmp paths and broad pkill patterns make some shell tests unsafe for concurrent runs.

  Rejected: 
  - ~~Reset recovery is incompatible with older peers while the protocol version remains unchanged.~~ I'm rejecting this one because this only matters when etserver restarts and comes up as the new version. Before this PR, etserver loses its keys and clients are dropped anyway. So in practice this can't be reached.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -475,6 +475,8 @@ add_library(
   src/terminal/SshSetupHandler.cpp
   src/terminal/UserTerminalHandler.hpp
   src/terminal/UserTerminalHandler.cpp
+  src/terminal/SessionStore.hpp
+  src/terminal/SessionStore.cpp
   src/terminal/UserJumphostHandler.hpp
   src/terminal/UserJumphostHandler.cpp
   src/terminal/TelemetryService.hpp
--- a/proto/ET.proto
+++ b/proto/ET.proto
@@ -19,6 +19,7 @@ enum ConnectStatus {
   RETURNING_CLIENT = 2;
   INVALID_KEY = 3;
   MISMATCHED_PROTOCOL = 4;
+  RETRY_LATER = 5;
 }
 
 message ConnectResponse {
@@ -28,6 +29,10 @@ message ConnectResponse {
 
 message SequenceHeader {
   optional int32 sequenceNumber = 1;
+  // Set when either side's sequence state is unusable (fresh process on one
+  // side). Both peers then zero their BackedReader/BackedWriter state and
+  // exchange empty catchup buffers instead of replaying history.
+  optional bool reset = 2;
 }
 
 message CatchupBuffer {
--- a/src/base/BackedReader.cpp
+++ b/src/base/BackedReader.cpp
@@ -104,6 +104,15 @@ void BackedReader::revive(int newSocketFd,
   socketFd = newSocketFd;
 }
 
+void BackedReader::reset() {
+  // Caller must hold recoverMutex (see getRecoverMutex), mirroring revive().
+  localBuffer.clear();
+  partialMessage.clear();
+  sequenceNumber = 0;
+  // Restart the stream nonce so a fresh peer process stays in lockstep.
+  cryptoHandler->resetNonce();
+}
+
 int BackedReader::getPartialMessageLength() {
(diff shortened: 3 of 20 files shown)


MisterTea:
@epsalmond Why close this?

epsalmond:
@MisterTea is was vibe-opened and wasn't ready.

I've since made a ton of improvements, it actually works as intended, and it's my full-time terminal workflow. If you're interested in it I'll re-open it with a month worth's of updates and bugfixes.

MisterTea:
Sure, go for it

Kronuz:
Nice work on this PR. Restart survival and the reset handshake are real gaps for sessions. Just flagging one thing: #819 also adds a client-side `--attach NAME` under `~/.et/sessions/` (written before I'd seen this PR), so the flag name and directory overlap with what you've built here.

The one in #819 is a much thinner piece (just a cached reconnect id/passkey with no restart survival and no title capture), and is needed for `etctl` control plane (#820). When this PR lands I'd like to rework the named sessions there onto your more complete `SessionStore`.

epsalmond:
@Kronuz Thanks! It has been battle-tested.

I had to stand up a windows vm to work through some of the stuff that has landed since I first opened this, and today there are a few merge conflicts. I expect I'll be ready to merge today.

epsalmond:
@MisterTea @Kronuz I think this is ready. 

Freebsd test failed this run but it looks flakey. PR is updated. Protocol backwards-compatible (so no bump) and I cleaned up the commits and comments. Rebased on top of #850 and #855.

Two bugfixes for master included: a terminal that re-registers after a restart never started the #855 disconnect deadline, and the #850 positional-command parser lost the host after the jumphost flag.

I don't have anything for the control plane in this PR but I'm happy to do a follow up on top of #819 to add `etserver --list`.

MisterTea:
@Kronuz if this PR lands, what's the state of your PRs?  Would you rebase on this?

psennats:
I want this too, let me know how I can be of any help.

Issue #787, opened 2026-08-02
etserver retains ~1.9 GiB after repeated failed recovery attempts


j2h4u:
## Summary

On an Eternal Terminal 7.0.0 server, a set of sessions left running overnight entered a failed recovery loop. Existing clients appeared frozen, and a fresh client failed with:

```
Error connecting to server: 3: Client is not registered
```

At the same time, `etserver` retained about 1.9 GiB of anonymous memory. Closing all ET clients stopped the retry storm and allowed a fresh connection. Restarting only `et.service` released the memory; the tmux server and all tmux sessions were unaffected.

## Environment

- Eternal Terminal: 7.0.0
- Server: Linux, systemd-managed `et.service`
- 5 server-side ET terminal states
- Interactive work was running inside tmux
- Clients had been left connected overnight

## Observed evidence

Before restarting `etserver`:

- approximately 824 MiB RSS plus 1.1 GiB swap
- `/proc/<pid>/smaps` showed 26 large anonymous mappings of roughly 64 MiB each
- about 1.69 GiB resident + swapped data was accounted for by those anonymous mappings
- server logs contained 199 failed recovery attempts and 13 successful recoveries during the incident
- a stale client ID kept retrying recovery while new clients received `Client is not registered`

After restarting only `et.service`:

- `etserver` returned to approximately 19 MiB RSS and 0 swap
- tmux PID/start time did not change
- reconnecting ET clients worked normally

## Potentially relevant code paths

I am not claiming a confirmed root cause, but these paths appear relevant on current master:

- `BackedWriter` can retain up to 64 MiB of recovery data plus up to 64 MiB accumulated while disconnected.
- `BackedWriter::recover()` serializes packets into a new `vector<string>`.
- `Connection::recover()` copies those strings into a protobuf `CatchupBuffer`, which is then serialized again.
- returning-client recovery is performed while `ServerConnection::classMutex` is held.

Repeated failed recovery may therefore produce large temporary allocations and allocator high-water retention, while also delaying unrelated client handling.

## Expected behavior

- Failed recovery attempts should not leave `etserver` retaining memory proportional to repeated catch-up buffers.
- One stale/retrying client should not prevent fresh clients from registering or reconnecting.

## Actual behavior

- Existing clients remained frozen.
- Fresh clients received `Client is not registered`.
- The server retained about 1.9 GiB until it was restarted.

## Reproduction status

[...]

Issue #782, opened 2026-07-28
Feature request: show a connection-status indicator while the client is disconnected / reconnecting


ttarabula:
**Component:** etterminal / et client
**Version:** behavior described against master @ 90711ad42 (et-v7.0.0); same on 6.2.8
**Type:** UX / feature request (no protocol change required for the core ask)

---

Thanks for ET — I love using it and it's part of my daily workflow. This is a nice-to-have, not a
must; sharing it because it would have turned a genuinely stressful experience into a non-event,
and it looks achievable with state the client already has.

## Problem

When an ET client loses contact with the server, it enters a background reconnect loop and keeps
retrying — but it gives the user **no visual indication** that this is happening. The terminal
simply freezes. From the user's seat, "disconnected but actively reconnecting, and likely to
recover if I wait" is **visually indistinguishable** from "permanently hung, kill it." That
ambiguity is the entire problem: without feedback, a recoverable stall reads as a dead session.

Concretely, I recently had a client frozen for ~2 hours during a server-side stall. The session
*would* have recovered on its own (and eventually did), but nothing on screen said so — no
"still trying," no "N seconds since last contact." Had there been even a one-line "reconnecting,
last contact Xs ago" indicator, I would have known to wait instead of assuming total failure.

## The client already has all the state — it just never renders it

Nothing new needs to be computed or added to the protocol. The client's main loop keeps running
throughout a disconnect and already tracks everything an indicator would need:

- **Disconnected state:** `clientFd < 0` in the main loop (`TerminalClient.cpp:347`).
- **Actively reconnecting:** the reconnect thread loops once per second
  (`ClientConnection.cpp:104` "In reconnect loop", `:154` "Waiting to retry...",
  `:93` "Trying to reconnect").
- **Time since last contact:** derivable from the keepalive bookkeeping already in the loop
  (`TerminalClient.cpp:335-349`, `waitingOnKeepalive`).
- **A render path to the user's terminal** already exists: `console->write(...)`
  (`TerminalClient.cpp:331`).

Today all of the above is emitted only via `LOG`/`VLOG` to the server/client log files, never to
the user's terminal. The only things ET writes to the screen are session bytes, a one-time
"ET running, feel free to background..." (`TerminalClient.cpp:180`), and terminal
`Connection closing.../Session terminated` messages (`:383`, `:392`).

## Prior art: mosh

mosh solves exactly this. On loss of contact it shows an overlay line such as:

[...]

PR #797, opened 2026-08-27
Rebase PR #730: opt-in backpressure/discard flow control

xz-dev:
This follow-up carries Ben Maurer's three-commit PR #730 stack rebased onto current `MisterTea/EternalTerminal@b74a12ef` (`master` as of 2026-08-06). Original authorship and commit order are preserved.

Upstream PR: #730
Related issues: #631, #787, #782

Why this matters beyond interactive lag: we reproduced a recovery deadloop under fast terminal output. With the default `none` mode, stale output grew into a multi-megabyte queue; reconnect repeatedly hit the recovery timeout and restarted from the beginning. In our exact v7 baseline reproduction, recovery never finished. A long-running tmux/Pi session on a lossy, high-RTT path can therefore appear permanently frozen even though the remote program and `etserver` remain alive. This matches the failed-recovery class reported in #787; #782 explains why users only see a frozen terminal while reconnect/recovery progress exists only in logs.

With this patch and explicit `--flow-control backpressure`, matched-binary tests bounded the queue, stalled the producer while congested, resumed it after recovery, and returned display lag to about 1–2 seconds. `discard` also recovered while allowing producer progress, with the documented scrollback trade-off. Default behavior remains unchanged unless a client opts in.

Validation on the current rebased branch:

- `cmake -S . -B build -DDISABLE_VCPKG=ON -DBUILD_TESTING=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo`
- `cmake --build build -j1`
- `ctest --test-dir build --output-on-failure -j1`: 159/159 passed, 2 privilege-dependent skips
- clang-format 18.1.8 dry-run: passed
- `git diff --check origin/master..HEAD`: passed
- repeated bounded E2E comparisons: `none` timed out; `backpressure` and `discard` recovered
- production canary against an existing v7 server: `backpressure` recovered without restarting the server
- real Tailscale-path canary: exact ~10 s outage, producer stall/resume, no `Socket Timeout`, display caught up below 2 s

The only rebase conflict was in `TerminalClient::run`; resolution preserves both PR #730 output-buffer draining and current master’s #788 non-TTY guard (`console && consoleFd >= 0`).

This is intended to help the original PR move forward, not replace its author. If Ben can update #730 directly, this branch can be used as a tested rebase reference or cherry-picked as the same three-commit stack.

### DIFF
--- a/proto/ETerminal.proto
+++ b/proto/ETerminal.proto
@@ -57,10 +57,25 @@ message PortForwardData {
   optional bool closed = 5;
 }
 
+enum FlowControlMode {
+  // Legacy behavior: no application-level buffering or kernel buffer
+  // tuning. This is the default and what servers assume for old clients
+  // that don't send flow_control_mode: current users are unchanged.
+  FLOW_CONTROL_NONE = 0;
+  // Bounded buffering; when full, stop reading from the source. Lossless
+  // (like plain ssh, but with small, tuned buffers so Ctrl-C stays
+  // responsive on a saturated link).
+  FLOW_CONTROL_BACKPRESSURE = 1;
+  // Bounded buffering; when full, drop the oldest pending output. The
+  // producing process never stalls and the display stays near real time.
+  FLOW_CONTROL_DISCARD = 2;
+}
+
 message InitialPayload {
   optional bool jumphost = 1 [default = false];
   repeated PortForwardSourceRequest reversetunnels = 2;
   map<string, string> environmentvariables = 3;
+  optional FlowControlMode flow_control_mode = 4 [default = FLOW_CONTROL_NONE];
 }
 
 message InitialResponse {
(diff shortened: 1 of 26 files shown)


xz-dev:
Validation context and linked production failure report are also posted on the original PR: https://github.com/MisterTea/EternalTerminal/pull/730#issuecomment-5435525546. This follow-up should be treated as a tested rebase aid for #730.

xz-dev:
Superseded by the focused follow-up PR against Ben Maurer's branch: https://github.com/bmaurer/EternalTerminal/pull/1

Closing this PR because the original contribution should remain primary. The focused follow-up preserves Ben's existing commits/authorship, merges current upstream, resolves the #788 conflict, and fixes the existing E2E cleanup bug without introducing another upstream-facing copy of PR #730.

PR #800, opened 2026-08-31
Document Mobile SSH as a third-party mobile ET client

dkozlov:
The README currently documents desktop/WSL clients but does not point mobile users to a native Android or iOS option. This adds a short, explicitly third-party note for Mobile SSH and labels both distribution channels as beta.

Disclosure: I maintain Mobile SSH. Its ET transport is optional; the app otherwise operates as an independent SSH client.


MisterTea:
@dkozlov please circle back when you are out of closed-beta and your app is available to the public, thanks.

MisterTea:
closing for now

PR #803, opened 2026-09-01
Fix terminal size query failure propagation

xz-dev:
## Summary

- make terminal-size query failure explicit in the internal `Console` contract
- return no observation when Unix `TIOCGWINSZ` fails
- keep the last successful geometry and suppress `TERMINAL_INFO` emission on failure
- add an end-to-end regression for success, failure, and unchanged recovery

## Why

`TIOCGWINSZ` can fail without writing the caller-provided `winsize`. The previous code ignored the return value, read that untouched storage, and could serialize it as remote terminal geometry.

This change handles the confirmed latent defect without adding receiver-side size policy, protocol changes, or arbitrary geometry limits.

## Validation

- `cmake --fresh -S . -B build -G Ninja`
- `ninja -C build`
- `./build/et-test TerminalInfoQueryFailure --reporter compact` — passed, 3 assertions
- deliberate failure injection converted absence to default geometry — regression failed at the receiver resize-count assertion; restored implementation passed
- `bash format.sh`
- `git diff --check`
- independent final review — approved with explicit residual risk

Full `ctest --parallel --output-on-failure` completed with three HTM system-test timeouts:

- `htm_pty_e2e`: `INSERT_KEYS` stalled while panes were printing
- `htm_features_e2e`: timed out waiting for concurrent `APPEND_TO_PANE`
- `htm_stress_e2e`: timed out waiting for bulk output and injected keys

These are likely unrelated by source reachability: the HTM tests do not launch `TerminalClient`, HTM does not call `getTerminalInfo()`, and the failures occur in HTM key/output flow. No clean `master` A/B run was recorded, so this is not claimed categorically. Two additional parallel integration timeouts (`EndToEndTest`, `JumphostEndToEndTest`) passed on serial/isolated reruns.

Native Windows and opt-in Ghostty, iTerm2, and Hyper E2E suites were not run.

### DIFF
--- a/src/terminal/Console.hpp
+++ b/src/terminal/Console.hpp
@@ -1,6 +1,8 @@
 #ifndef __CONSOLE_HPP__
 #define __CONSOLE_HPP__
 
+#include <optional>
+
 #include "ETerminal.pb.h"
 #include "Headers.hpp"
 #include "RawSocketUtils.hpp"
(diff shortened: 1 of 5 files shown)


MisterTea:
Thanks!

PR #809, opened 2026-09-09
Replace select() with epoll/kqueue and poll()

splhack:
select() writes past the end of an fd_set once a descriptor reaches FD_SETSIZE, corrupting the stack, e.g. SIGSEGV. `et -t 8000-9100:8000-9100` reaches that from the command line alone, since each forwarded port is its own listening descriptor.

Add FdPoller (epoll on Linux, kqueue on BSD/macOS) for the server loops, use poll() for single-descriptor waits and the client loop, and process only the descriptors reported ready. Windows keeps select(), where fd_set is a bounded array and cannot overflow. The client uses poll() rather than FdPoller because nohup(1) can leave a regular file on the console descriptor, which epoll_ctl rejects with EPERM.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -428,6 +428,10 @@ add_library(
   src/base/Connection.cpp
   src/base/CryptoHandler.hpp
   src/base/CryptoHandler.cpp
+  src/base/FdPoller.hpp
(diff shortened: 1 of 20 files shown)


MisterTea:
Thanks!

PR #810, opened 2026-09-10
Fix session and forwarding FD leaks

splhack:
Close completed sessions' router connections and remaining client/server forwarding sockets. Remove generated socket directories. Preserve reconnects and polling timeouts.

Validation: macOS build, all 282 CTest tests, and `bash format.sh` (`clang-format` 18).


splhack:
added more tests for coverage

MisterTea:
Thanks!

PR #811, opened 2026-09-10
Honor explicit TCP forwarding destinations

msokolov-meta:
TCP forwarding ignores the destination hostname and connects to loopback, even when the tunnel argument specifies an explicit host. This can fail to reach the requested service or connect to a different service on loopback at the same port.

Pass non-empty TCP destinations to `TcpSocketHandler::connect` unchanged. Missing or empty names retain IPv6-first loopback forwarding with IPv4 fallback. An explicit destination that fails to connect returns an error; it does not fall back to loopback.

Locking and connection timeouts are unchanged. Slow lookups or connection attempts can still stall other sessions sharing the socket handler. The wire protocol and UNIX socket forwarding are unchanged.

Update the existing forwarding tests and document the existing four-part tunnel syntax.

## Testing

- Linux Debug build and default CTest suite: 273/273 passed, including the headless terminal E2E tests.
- Existing real-socket harness in an isolated network namespace: 20/20 baseline/patched checks passed. Explicit IPv4, IPv6, and named destinations reached the selected host instead of a loopback decoy. Legacy fallback, refused and unresolved destinations, and 1 MiB round trips passed. The baseline reproduced the destination bug.
- Changed-file formatting and `git diff --check` passed.

Native macOS and Windows validation of this upstream tree has not been run locally.


msokolov-meta:
Historical validation for `c6b0d6686a33`, before the destination-only update:

- Linux Debug build and default CTest suite: 273/273 passed, including the headless terminal E2E tests.
- Formatting and `git diff --check` passed.
- A standalone real-socket harness in a private network namespace compared the patch with unmodified upstream. Explicit IPv4, IPv6, and named destinations reached the selected host instead of a loopback decoy. Legacy loopback behavior, failed destinations, and 1 MiB round trips passed.
- During a three-second stalled connect, an established socket's round trip took about 2.85 seconds upstream and 0.08–0.10 ms with the patch across three trials.
- Eight-thread connection churn completed 800 successful and 800 refused connections per build, with no observed descriptor leaks.

Native macOS and Windows validation has not been run locally.

The earlier latency improvements depended on the removed lock-narrowing change and no longer apply. Current validation is recorded in the PR description.

MisterTea:
Thanks!

PR #812, opened 2026-09-11
Add optional raw stack traces

splhack:
Synchronous symbolization can hold the logger mutex and block new connections. Add `RAW_STACKTRACE` (default OFF) to log PCs, module paths, and load addresses for offline symbolization.

Tested: macOS build, 270 tests, formatting, and forwarding-reset repro with offline symbolization.

----

I observed etserver freeze completely for about 10 minutes on two separate servers while waiting for addr2line to symbolize stack traces. During each freeze, new connections could not be established, and all existing sessions became unresponsive.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,6 +222,7 @@ endif()
 option(CODE_COVERAGE "Enable code coverage" OFF)
 option(FUZZING "Enable builds for fuzz testing" OFF)
 option(DISABLE_CRASH_LOG "Disable installing easylogging crash handler" OFF)
+option(RAW_STACKTRACE "Log stack traces without symbolization" OFF)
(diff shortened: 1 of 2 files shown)


jshort:
Rerunning CI jobs due to timeout.

MisterTea:
This PR turns off symbolication or does the PR make it asynchronous?  Could we make it asynchronous or will that cause issues?

splhack:
> This PR turns off symbolication or does the PR make it asynchronous? Could we make it asynchronous or will that cause issues?

This PR turns off symbolication when build time option `RAW_STACKTRACE` is set.

Asynchronous symbolication might be possible on top of that
((1) preserve raw stacktrace (2) symbolicate it asynchronously.)

PR #815, opened 2026-09-17
Preserve partial socket write progress

miaoever:
## Summary

- preserve the number of bytes already committed by `UnixSocketHandler::write()` when a later write times out or encounters a hard error
- use `ssize_t` for native write results and accumulated byte counts
- add regression tests for both the `EAGAIN` timeout and hard-error cases

## Problem

`UnixSocketHandler::write()` loops internally and may successfully write a prefix before later calls fail. It previously returned `-1` in that case, hiding the committed prefix from callers such as `writeAllOrThrow()`.

During recovery, the caller then retries from its unchanged offset and sends that prefix twice. This corrupts the length-framed `CatchupBuffer` protobuf and the peer reports `Invalid proto`. Large catch-up buffers after a long disconnect make this failure much easier to trigger.

Returning the positive partial count preserves normal `write(2)` semantics, allowing the outer write loop to resume at the correct offset. A failure before any bytes are committed still returns `-1`.

## Testing

- `./build/et-test '[UnixSocketHandler]' --reporter compact` — 7 test cases, 47 assertions passed
- `ctest --test-dir build --parallel 8 --output-on-failure` — 292/292 passed (3 opt-in GUI/PTY tests skipped)
- formatted changed files with `clang-format --style=Google`


src/base/UnixSocketHandler.cpp
```diff
@@ -83,16 +83,16 @@ ssize_t UnixSocketHandler::write(int fd, const void* buf, size_t count) {
         std::this_thread::sleep_for(std::chrono::milliseconds(1));
         if (time(NULL) > startTime + 5) {
           // Give up
-          return -1;
+          return bytesWritten > 0 ? bytesWritten : -1;
         }
       } else {
-        return -1;
+        return bytesWritten > 0 ? bytesWritten : -1;
```
MisterTea, line 89: This looks dangerous: there could be code that expected to write everything or get a -1.  We'll have to make sure everyone who calls this expects partial writes.

But I trust CI, if CI passes, we'll land this and see what happens :-)

PR #822, opened 2026-09-20
fix: show reconnect status once and clear it on recovery (#782)

### LINKED ISSUE #782: Feature request: show a connection-status indicator while the client is disconnected / reconnecting

MisterTea:
Emits one stable reconnect notice per outage and clears it after recovery using terminal-safe ASCII. Replaces root triage notes and an invalid local test with focused tests; 4 targeted assertions pass.

Closes #782.


MisterTea:
Closing this PR and issue #782.

After technical evaluation, rendering an in-band reconnect indicator directly to the client terminal is unsafe for ET's architecture:

1. **No Client-Side Framebuffer (Why Mosh can do this, but ET cannot):**
   Mosh implements a complete client-side terminal emulator and owns the screen matrix, allowing it to paint an overlay banner and restore damaged cells cleanly. ET is an end-to-end transparent PTY byte pipe; it does not parse or track the terminal screen state.

2. **Destructive to Curses/TUI Applications:**
   Full-screen applications (`vim`, `tmux`, `htop`, `nano`, `fzf`) manage the screen grid directly. Moving to the bottom row and clearing it on reconnect wipes out status bars and command lines without notifying the host application, leaving the screen corrupted until the user manually triggers a full redraw (`Ctrl+L`). Furthermore, cursor save/restore sequences (`\033[s` / `\033[u`) are not universally reliable across terminal emulators or when alternate screen buffers are active.

3. **Incompatible with tmux Control Mode (`tmux -CC` / `htm`):**
   When ET carries tmux control mode to emulators like iTerm2 or WezTerm, the client expects the line-oriented tmux control protocol (`%` notifications). Injecting raw ANSI escape sequences into the stream causes the emulator to treat the message as an unrecognized command or corrupted stream and immediately tear down tmux mode.

4. **Stream Splicing Risks:**
   Injecting status bytes into the console descriptor can cut across multi-byte UTF-8 sequences or partial escape sequences that were interrupted mid-transit.

Because ET cannot safely render an overlay into an unmanaged terminal stream without a full client-side terminal emulator, we cannot implement an in-band visual indicator without risking session corruption.

PR #824, opened 2026-09-20
docs: record named-session reattachment findings (#365)

### LINKED ISSUE #365: re-attach to session
How can I re-attach to a session if my local shell/computer crashes? When I use et again I get a new shell on the remote machine. A quick `w` shows my other session and job is still running.

I also noticed looking through previous issues before posting that only 1 et connection can run at a time (issue from 2017), but I have been running multiple. Maybe this has changed since that old issue.

MisterTea:
## Problem

See #365 for the original report.

## Changes

Records findings about named sessions, control-mode attach support, and stale-client takeover behavior.

## Resolution status

This is investigation material only. It does not add user documentation or change reattachment behavior, so it does not resolve #365.

## Verification


MisterTea:
Superseded by #793, which implements named-session reattachment, listing, last-heartbeat visibility, and session lifecycle behavior with real integration coverage. This placeholder findings-only PR does not add a second competing implementation.

Issue #428, opened 2021-04-21
[Request] way to list unattached etterminal's and date/time of last attachment


paulraines68:
After over a month of ET use on our ssh gateway server we have a situation of etterminal process like this parsed from the utmp data

```
user1     pts/20      9274 2021-03-29 10:18  et [9274]
user2     pts/28     19940 2021-03-29 12:27  et [19940]
user3     pts/19     15292 2021-03-30 13:20  et [15292]
user1     pts/25     28569 2021-03-30 16:05  et [28569]
user2     pts/29      5117 2021-03-31 13:48  et [5117]
user3     pts/30      5360 2021-03-31 13:50  et [5360]
user1     pts/26     23472 2021-04-01 13:21  et [23472]
user2     pts/32      7223 2021-04-09 12:04  et [7223]
user3     pts/14     24589 2021-04-09 15:04  et [24589]
user1     pts/18     29069 2021-04-12 14:09  et [29069]
user2     pts/27     17757 2021-04-14 11:02  et [17757]
user3     pts/33     18293 2021-04-14 11:08  et [18293]
user4     pts/37     28126 2021-04-15 14:52  et [28126]
user1     pts/34     20577 2021-04-16 09:39  et [20577]
user2     pts/40       803 2021-04-16 12:05  et [803]
user2     pts/41      1000 2021-04-16 12:06  et [1000]
user1     pts/42      8796 2021-04-16 13:42  et [8796]
user4     pts/43     19734 2021-04-16 15:52  et [19734]
user1     pts/8      15110 2021-04-17 11:08  et [15110]
```

Most of these are surely abandoned for good to never be used again (user just killed or rebooted their client machine without properly exiting from attached session).  How can I determine which of these are actively attached and for those unattached how long they have been unattached?  I would like to set up a cron job to kill all etterminal processes unattached for over a week.

gklanderman:
etterminal should probably just have a timeout and clean up temp files and exit if it doesn't hear from et in say 24 hours (configurable) rather than wait around forever.

seems like et could also do a better job of shutting down the remote etterminal when it exits: it should catch errors and signals and try to clean up the remote if possible.  AFAICT, when using the --no-terminal option, there's no way to kill et cleanly so that the etterminal gets cleaned up.

yobnoj:
[...]

PR #826, opened 2026-09-20
design: prototype a session registry for unattached sessions (#428)

MisterTea:
## Problem

See #428 for the original report.

## Changes

Introduces a header-only API declaration, design notes, and proposed tests for tracking attachment state and expiry.

## Resolution status

The registry methods have no implementation and the registry is not integrated with the server or a listing command. This branch does not yet provide the functionality requested in #428.

## Verification


MisterTea:
Superseded by #793, which implements the requested session listing and lifecycle behavior. This branch only declares an unintegrated registry, so closing it avoids presenting a non-functional design stub as a fix.

Issue #708, opened 2025-10-27
Error binding port when port forwarding even though existing port is outgoing


p12tic:
Saw the following problem when trying to connect with port forwarding (`-t 12340-12350:12340-12350`):

```
> et -p 10000 <redacted> -t 12340-12350:12340-12350
Enter passphrase for key '<redacted>': 
Error binding 2/1/6: 98 Address already in use

Error establishing port forward: Error binding port 12344: 98 Address already in use
```

`sudo netstat -lnpt | grep 12344` produces nothing, means nothing is listening on that port

Only using lsof revealed the culprit:

```
sudo lsof -i tcp:12344
COMMAND  PID USER FD   TYPE  DEVICE SIZE/OFF NODE NAME
slack   ...  ... ...  IPv4 1013640      0t0  TCP <redacted>:12344->ec2-52-215-161-110.eu-west-1.compute.amazonaws.com:https (ESTABLISHED)
```

Restarting slack allowed me to connect via `et`.

Looks like et port forwarding conflicts with any forwarded port being used instead of only with ports that are being open for listening.

PR #838, opened 2026-09-20
fix: repair listener tuple handling for port forwarding (#708)

MisterTea:
## Problem

See #708 for the original report.

## Changes

This branch currently contains unrelated SSH-banner and comma-separated tunnel commits plus local, uncommitted listener experiments.

## Resolution status

The committed branch does not address #708 cleanly and should not be merged in its current form.

## Verification


MisterTea:
Closing this branch because its committed changes are unrelated to #708 and the local listener experiments do not yet establish a correct fix. The untracked investigation files remain preserved in the worktree. A follow-up should start from current master and include a Linux regression that reproduces the outgoing-connection/listener collision before changing socket reuse behavior.

PR #839, opened 2026-09-20
test: sketch FreeBSD reconnect logging coverage for #734

### LINKED ISSUE #734: FreeBSD 15.0-RELEASE session does not reconnect.
Eternal Terminal does not reconnect to a session on FreeBSD 15.0-RELEASE and fails with the message `Session terminated`.
FreeBSD server machine is running et version 6.2.11.

Client is Ubuntu 24.04 running et version 6.2.10.
I have another Ubuntu 24.04 machine that is also running 6.2.10 and I try to do the same exact use case and the session seems to reconnect just fine.

I am not sure if this is due to the point release difference or the different OS.

MisterTea:
## Problem

See #734 for the original report.

## Changes

Adds a proposed reconnect/verbose-log test.

## Resolution status

No reconnect behavior changes and the test is not integrated. This branch does not fix #734.

## Verification


MisterTea:
Closing because this branch contains no implementation or executable regression. Current master has since replaced the affected readiness layer with the poll/kqueue work from #809; a new FreeBSD 15 report with current-version verbose logs should be handled in a focused follow-up if the problem remains.

Issue #778, opened 2026-07-17
Unfixed CVE-2023-23558


abergmann:
I've noticed that Eternal Terminal [GHSA-m7cp-w3xh-qr3p ](https://github.com/advisories/GHSA-m7cp-w3xh-qr3p) is still in state `Unreviewed`.

There was an [issue#560](https://github.com/MisterTea/EternalTerminal/issues/560) that mentioned CVE-2023-23558, but the question if the issue was fixed or not was never answered.

Could you please have a look and comment here?

abergmann:
The predictable temporary dir path is still there:

https://github.com/MisterTea/EternalTerminal/blob/3dd946d7128ea98653bbbab2f454706aa66d9893/src/terminal/TerminalClientMain.cpp#L224

MisterTea:
The CVE isn't of any concern to me. There's nothing so sensitive in telemetry that another user on the machine shouldn't be able to see it. Feel free to submit a PR if you want to fix it.

PR #844, opened 2026-09-20
security: audit per-user temporary path handling for #778

MisterTea:
## Problem

See #778 for the original report.

## Changes

Adds audit notes covering telemetry configuration, FIFO paths, PID-file permissions, and possible symlink races.

## Resolution status

The files are documentation/placeholders rather than executable security tests, and no hardening change is included. This PR does not establish that CVE-2023-23558 is fixed.

## Verification


MisterTea:
Superseded by the security hardening and regression coverage merged in #784. This branch contains audit notes and non-executable placeholders only, so it should not remain open as a competing security fix.

Issue #707, opened 2025-10-24
Feature request: cleanly shut down remote side when closing local terminal window


th-in-gs:
I have a terrible habit of just closing terminal windows with I’m done with them, even if I’m `et`-ed to another server.

It would be great if `et` could (optionally?) catch `SIGHUP` and close the connection cleanly when I do this, so that there are not dangling `etterminal` processes left on the server.

PR #846, opened 2026-09-20
fix: cap per-session recovery memory (#787)

### LINKED ISSUE #787: etserver retains ~1.9 GiB after repeated failed recovery attempts

MisterTea:
Reduces recovery history and disconnected buffering from 64 MiB each to 8 MiB per connection while preserving replay behavior within the bounded window. Removes the empty placeholder test and preserves unrelated local edits.

Closes #787.


MisterTea:
Closing after investigation of #787:

1. The reported ~1.9 GiB memory footprint with 26 anonymous mappings of ~64 MiB in `/proc/<pid>/smaps` corresponds to sessions reaching the 64 MiB recovery/disconnected buffer limit in `BackedWriter` (`MAX_BACKUP_BYTES` and `DISCONNECT_BUFFER_BYTES`).
2. Reducing the buffer from 64 MiB to 8 MiB degrades recovery and disconnected output buffering capabilities and is not viable, so that change has been reverted.
3. The root operational cause of unbounded buffer retention is abandoned/stale sessions remaining alive indefinitely. PR #837 introduces `--close-on-hangup` (#707), allowing users to configure sessions to terminate on client exit, naturally releasing sessions and their memory buffers.

PR #837 closes #787 as well.

